PR #1, opened 2019-09-04
Add a "meta" RFC for our RFC process!

DFurnes:
This pull request introduces the "RFCs" RFC:

> We want to start using RFCs to discuss (and document) architectural decisions.

## Open Questions
- How do folks feel about the template? I wanted to keep it as lightweight as possible, while also giving a clear outline for things that are worth mentioning in a "good" RFC. Anything I missed?
- How do folks feel about the suggested process (see `README.md`)?

### DIFF
--- a/001-rfcs.md
+++ b/001-rfcs.md
@@ -7,7 +7,7 @@ We want to start using RFCs to discuss (and document) architectural decisions.
 
 ## Problem
 
-Ideas tend to get stuck in one person's head by default (or lost in 1:1 conversations). As our team grows, we want to make it easy for everyone to take part in architectural discussions & see where we're thinking about taking our platform.
+Ideas tend to get stuck in one person's head by default (or lost in 1:1 conversations). As our team grows, we want to make it easy for everyone to take part in architectural/process discussions & see ideas we've committed to trying.
 
 ## Proposal
 
--- a/001-rfcs.md
+++ b/001-rfcs.md
@@ -3,7 +3,7 @@
 
 We want to start using RFCs to discuss (and document) architectural decisions.
 
-**Discussion:** Link the PR introducing this RFC here!
+**Discussion:** https://github.com/DoSomething/rfcs/pull/1
 
 ## Problem
 
--- a/001-rfcs.md
+++ b/001-rfcs.md
@@ -13,7 +13,7 @@ Ideas tend to get stuck in one person's head by default (or lost in 1:1 conversa
 
 We should adopt a simple RFC process, inspired by the process used by the engineering teams at [Artsy](https://artsy.github.io/blog/2019/04/11/on-an-rfcs-process/) and [GOV.UK](https://github.com/alphagov/govuk-rfcs) and on larger open-source projects like [React](https://github.com/reactjs/rfcs) or [Rust](https://github.com/rust-lang/rfcs). (And while their namesake comes from the [exhaustive IETF standards documents](https://www.rfc-editor.org/rfc-index.html), we definitely don't want to replicate that!)
 
-By documenting architectural goals in a semi-formal manner like this, we hope to increase transparency (since a public document is by definition a lot more transparent than one-one-one discussions). We're also hoping this reduces friction by helping to spread ideas _before_ the pull request review process.
+By documenting architectural goals in a semi-formal manner like this, we hope to increase transparency (since a public document is by definition a lot more transparent than 1:1 discussions). We're also hoping this reduces friction by helping to spread ideas _before_ the pull request review process.
 
 I'd recommend we start with a [very lightweight process](https://github.com/DoSomething/rfcs/tree/hello-rfcs#process) and introduce more rules over time.
 
--- a/001-rfcs.md
+++ b/001-rfcs.md
@@ -11,7 +11,7 @@ Ideas tend to get stuck in one person's head by default (or lost in 1:1 conversa
 
 ## Proposal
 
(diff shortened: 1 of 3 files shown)


mirie:
I think the template is great! 

In terms of process, I like the simplicity as well. I have two questions:
1. For this step: `Create a pull request and drop a link in #announce-tech so others can review it.` -- so today you put your request in the standup thread. Is that the place to call out a link to the RFC PR, or is the thinking to drop the link outside of that thread? Do we want to tag people (if we are looking for specific opinions) on the PR OR in the slack message (or both)?
2. What's the expectation we want to set for having some eyeballs on the RFC? Are we hoping for review and _initial_ feedback within say, 5 business days? or less time / more time?

001-rfcs.md
```diff
@@ -0,0 +1,25 @@
+
+# RFCs
+
+We want to start using RFCs to discuss (and document) architectural decisions.
+
+**Discussion:** https://github.com/DoSomething/rfcs/pull/1
+
+## Problem
+
+Ideas tend to get stuck in one person's head by default (or lost in 1:1 conversations). As our team grows, we want to make it easy for everyone to take part in architectural discussions & see where we're thinking about taking our platform.
```
katiecrane: I see this mainly referring to architectural discussions, but I know we've talked about using RFCs for how our team operates and other non-architectural processes. Would it make sense to update the wording here to show that the suggestions don't have to be architectural?

DFurnes:
> For this step: `Create a pull request and drop a link in #announce-tech so others can review it.` -- so today you put your request in the standup thread. Is that the place to call out a link to the RFC PR, or is the thinking to drop the link outside of that thread?

I think just dropping it in the channel (or even adding notifications for new PRs created in this repo) seems fine. I just called out in the standup thread to make sure it didn't get lost at the end of the day!

>  Do we want to tag people (if we are looking for specific opinions) on the PR OR in the slack message (or both)?

Definitely!

> What's the expectation we want to set for having some eyeballs on the RFC? Are we hoping for review and _initial_ feedback within say, 5 business days? or less time / more time?

Hmm, I'm not sure! Let's see how things go and perhaps add some guidelines in the future?

DFurnes:
I've added a "historical" RFC for the [environment naming rules](https://github.com/DoSomething/rfcs/blob/master/002-resource-naming.md) we standardized last year, and another example RFC for continued GraphQL investment in https://github.com/DoSomething/rfcs/pull/3!

001-rfcs.md
DFurnes: Mmmm, good point! I tried re-phrasing this in 8baebe7.

katiecrane (approved):

001-rfcs.md
```diff
@@ -0,0 +1,25 @@
+
+# RFCs
+
+We want to start using RFCs to discuss (and document) architectural decisions.
+
+**Discussion:** https://github.com/DoSomething/rfcs/pull/1
```
mendelB, line 6: I really dig the meta-ness of this self-reference, within the meta RFC for the RFC! ♻️

mendelB (approved):
👏 Yes Yes Yes! This makes great sense!

weerd (approved):
🎉

weerd:
Also, thoughts on having a directory in the repo that contains all the RFCs instead of top level? Just for organization purposes.

DFurnes:
> Also, thoughts on having a directory in the repo that contains all the RFCs instead of top level? Just for organization purposes.

Since there's only two non-RFC files (that are really just for GitHub to parse), I think I'd rather avoid making people take an extra click to see the list. Can always re-evaluate if this repo gets messier!

DFurnes:
Rebased to fix a merge conflict with the README. 💥

001-rfcs.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Problem
+
+Ideas tend to get stuck in one person's head by default (or lost in 1:1 conversations). As our team grows, we want to make it easy for everyone to take part in architectural/process discussions & see ideas we've committed to trying.
+
+## Proposal
+
+We should adopt a simple RFC process, inspired by the process used by the engineering teams at [Artsy](https://artsy.github.io/blog/2019/04/11/on-an-rfcs-process/) and [GOV.UK](https://github.com/alphagov/govuk-rfcs) and on larger open-source projects like [React](https://github.com/reactjs/rfcs) or [Rust](https://github.com/rust-lang/rfcs). (And while their namesake comes from the [exhaustive IETF standards documents](https://www.rfc-editor.org/rfc-index.html), we definetely don't want to replicate that!)
```
mshmsh5000: _definetely_ ➡️ _definitely_

001-rfcs.md
```diff
@@ -0,0 +1,25 @@
...
+
+By documenting architectural goals in a semi-formal manner like this, we hope to increase transparency (since a public document is by definition a lot more transparent than one-one-one discussions). We're also hoping this reduces friction by helping to spread ideas _before_ the pull request review process.
```
mshmsh5000: Small-potatoes, but I'd stick with either _1:1_ or _one-on-one_ across the doc, just so it's easier to scan.

mshmsh5000 (approved):
Love it! Just two small language comments.

rapala61:
Excited to start using RFCs! Great work 💯 .

rapala61 (approved):

001-rfcs.md
DFurnes: Ahh! Good catch. 🦅

001-rfcs.md
DFurnes: Agreed! Changed to use "1:1" here as well.

DFurnes:
Thanks everyone for the feedback! I'm going to merge this in! 🚀

PR #2, opened 2019-09-05
Add Resource Naming RFC.

DFurnes:
This pull request adds the (previously discussed) "Resource Naming" RFC:

> We should consistently name our resources (applications, databases, URLs, etc).

# Open Questions
N/A! This is an example RFC, based on an [discussion](https://github.com/DoSomething/internal/issues/366) we had last year.

### DIFF
--- a/002-resource-naming.md
+++ b/002-resource-naming.md
@@ -1,0 +1,33 @@
+
+# Resource Naming
+
+We should consistently name our resources (applications, databases, URLs, etc).
+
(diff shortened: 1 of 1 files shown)


DFurnes:
Since this was previously discussed in [DoSomething/internal#366](https://github.com/DoSomething/internal/issues/366), I'm just going to merge this in.

weerd (commented):
👍

PR #3, opened 2019-09-05
Add a RFC for continued GraphQL investment.

DFurnes:
This pull request introduces the ["GraphQL" RFC](https://github.com/DoSomething/rfcs/blob/graphql-all-the-things/003-graphql.md):

> We should use GraphQL for all data-fetching in the future.

# Open Questions
- Does this make a clear case for why we've been standardizing data fetching with GraphQL?
- Is there anywhere we _wouldn't_ want to use GraphQL? (e.g. where REST is better?)


003-graphql.md
```diff
@@ -0,0 +1,36 @@
...
+
+This gives us a single [universal schema](https://principledgraphql.com/integrity#1-one-graph) (so developers don't need to worry about which application holds which data). It lets us optimally [batch requests to backend services](https://www.apollographql.com/docs/graphql-tools/connectors/#dataloader-and-caching) and acts as a centralized "wrapper" around third-party APIs like [Contentful](https://www.contentful.com/).
+
+It also allows us to organically evolve our API over time (since we can add fields without bloating existing requests that don't need them, and [deprecate old fields](https://www.apollographql.com/docs/graphql-tools/generate-schema/#descriptions--deprecations) without breaking existing clients). We also get fantastic developer tools, like the [Apollo Chrome extension](https://chrome.google.com/webstore/detail/apollo-client-developer-t/jdkknkkbebbapilgoeccciglkfbmbnfm?hl=en-US) and [Visual Studio Code extension](https://marketplace.visualstudio.com/items?itemName=apollographql.vscode-apollo).
+
+We should continue to roll out GraphQL and reduce our reliance on custom data-fetching logic!
+
+## Drawbacks/Alternatives
```
mendelB, line 28: Perhaps we can add as a Drawback the team's relative lack of GraphQL expertise vis-à-vis traditional REST? This could be compounded for more complex GraphQL schema/resolving, caching, or other issues specifically introduced or made more complex by GraphQL's technology (e.g. something like the sensitive/optional fields work in [DoSomething/graphql#128](https://github.com/DoSomething/graphql/pull/128)).

Practically speaking this could introduce a slower development cycle and a higher chance of bugs creeping in.

(I personally think these are offset by the benefits, but maybe worth including?)

003-graphql.md
```diff
@@ -0,0 +1,36 @@
...
+
+Using GraphQL relies on more tooling than traditional REST HTTP requests (since you'd likely want a purpose-build GraphQL client). It also means we're investing in a particular technical stack that may not always be as popular [as it currently is](https://www.apollographql.com/docs/intro/benefits/#ready-for-production) (whereas REST isn't going anywhere).
```
mshmsh5000: _purpose-build_ ➡️ _purpose-built_

DFurnes: Fixed in aae9503!

003-graphql.md
```diff
@@ -0,0 +1,36 @@
...
+ - Finally, we had to do a lot of busy-work to [normalize our Redux state](https://redux.js.org/recipes/structuring-reducers/normalizing-state-shape) in Phoenix, which slowed down development of new features. (See [DoSomething/phoenix-next#767](https://github.com/DoSomething/phoenix-next/pull/767) and [DoSomething/phoenix-next#1123](https://github.com/DoSomething/phoenix-next/pull/1123)).
+
+## Proposal
+
+We built our GraphQL gateway (see [graphql.dosomething.org](https://graphql.dosomething.org/)) in November 2017. We have gradually added more resources (such as Contentful spaces and embed metadata) and applications (such as Gambit) over time.
+
+This gives us a single [universal schema](https://principledgraphql.com/integrity#1-one-graph) (so developers don't need to worry about which application holds which data). It lets us optimally [batch requests to backend services](https://www.apollographql.com/docs/graphql-tools/connectors/#dataloader-and-caching) and acts as a centralized "wrapper" around third-party APIs like [Contentful](https://www.contentful.com/).
```
mendelB, line 22: > This gives us a single [universal schema](https://principledgraphql.com/integrity#1-one-graph) (so developers don't need to worry about which application holds which data).

Perhaps this (the issue of multiple API's needing to be stitched together awkwardly) can be stressed more as a pain point in the `Problems` section? Maybe added to the first point there? (I always saw this as one of the main reasons for our adoption, but didn't feel the argument strongly in the problems section, though that could just be a personal misunderstanding.)

mendelB:
> Is there anywhere we wouldn't want to use GraphQL? (e.g. where REST is better?)

One place where I'm curious to see how switching to GraphQL will pan out for fetching from Contentful on the Phoenix backend. For some of the requests, this should be straightforward, but for some of the top-level deeply resolved requests for e.g. a [Campaign page](https://github.com/DoSomething/phoenix-next/blob/b42097394b5a76011de8ffabc42e2038b47352c1/app/Http/Controllers/CampaignController.php#L61), this might highlight some issues 🤔 

Not saying REST is better in such a case, but it's just a scenario that occurs to me as _potentially_ problematic.

003-graphql.md
DFurnes: Yeah, I think that's a really good point. I tried to address this better in 350ea7e, but definitely open to further suggestions!

003-graphql.md
DFurnes: Yeah, that was definitely missing! Added as the first bullet point in b01c6f6.

DFurnes:
> One place where I'm curious to see how switching to GraphQL will pan out for fetching from Contentful on the Phoenix backend. For some of the requests, this should be straightforward, but for some of the top-level deeply resolved requests for e.g. a [Campaign page](https://github.com/DoSomething/phoenix-next/blob/b42097394b5a76011de8ffabc42e2038b47352c1/app/Http/Controllers/CampaignController.php#L61), this might highlight some issues 🤔
> 
> Not saying REST is better in such a case, but it's just a scenario that occurs to me as _potentially_ problematic.

I'd be curious to talk more about this! Our Contentful space can be a uniquely problematic citizen in GraphQL-land because of how flexible it lets us be, but I feel like GraphQL actually solves more problems than it introduces there (for example, eventually not having to load _all_ the content, sub-pages, and blocks for a campaign just to render the pitch page).

mendelB:
> I'd be curious to talk more about this! Our Contentful space can be a uniquely problematic citizen in GraphQL-land because of how flexible it lets us be, but I feel like GraphQL actually solves more problems than it introduces there (for example, eventually not having to load _all_ the content, sub-pages, and blocks for a campaign just to render the pitch page).

Hmm yeah, I totally feel you! I guess my 'highlight some issues' statement might be better phrased as 'require some thoughtful structural amendments' so that e.g. we _aren't_ fetching a big nested resolved blob from Contentful, and perhaps resolving more thing via GraphQL from the front end. 

So I guess, not really a good case of where we _wouldn't_ want to use GraphQL!

mendelB (approved):
Oops, I forgot to hit approve!

003-graphql.md
aaronschachter: Another benefit that's perhaps assumed by having a single universal schema, but maybe worth spelling out here is that there will only be a single source of authentication for the DS API. Gambit, for example, is an app that stores authentication credentials for GraphQL, Northstar (via [Northstar JS](5d83c1a93d6e6a009c355da2) -- I've long wanted to replace this package for GraphQL), and Rogue -- only because we haven't swapped our GET/POST requests for GraphQL equivalents.

Also - back in the old days when I was working on the Gambit+GraphQL integration, I was blocked by using GraphQL for users because of authenticating as an admin machine. Is that still the case? Ideally Gambit Admin would pull message history from GraphQL as well (but I never added Gambit messages into GraphQL because of the authentication issues)

003-graphql.md
```diff
@@ -0,0 +1,42 @@
...
+
+It also allows us to organically evolve our API over time (since we can add fields without bloating existing requests that don't need them, and [deprecate old fields](https://www.apollographql.com/docs/graphql-tools/generate-schema/#descriptions--deprecations) without breaking existing clients). We also get fantastic developer tools, like the [Apollo Chrome extension](https://chrome.google.com/webstore/detail/apollo-client-developer-t/jdkknkkbebbapilgoeccciglkfbmbnfm?hl=en-US) and [Visual Studio Code extension](https://marketplace.visualstudio.com/items?itemName=apollographql.vscode-apollo).
+
```
aaronschachter, line 25: I keep coming back to "Does this make a clear case for why we've been standardizing data fetching with GraphQL?" and thinking of more, and it's related to this section -- but it's pretty great that we have / will-have-once-everything-is-ported-over standard camelCase naming standard on all resources. Related to deprecating old fields is having easier control of the names of resources, e.g. many of the Gambit schema names don't match the Gambini content types, and although the neat freak in all of us would prefer the name to be same all over, it's questionable whether a migration to change the machine name is worth it (especially if we want to change it again and again)

aaronschachter: We also get built in documentation by using the Playground (again, maybe a given - but definitely helps with onboarding I think)

aaronschachter (approved):
Left some ideas on benefits of GraphQL, feel free to take or leave. Looks good! Hope we can eventually fully replace all of Gambit's northstar-js and gateway-js requests with GraphQL ones.

003-graphql.md
DFurnes: > Also - back in the old days when I was working on the Gambit+GraphQL integration, I was blocked by using GraphQL for users because of authenticating as an admin machine. Is that still the case? Ideally Gambit Admin would pull message history from GraphQL as well (but I never added Gambit messages into GraphQL because of the authentication issues)

Yeah, if I remember the issue correctly, this is still the case. GraphQL "supports" Northstar's OAuth tokens for Northstar & Rogue simply because it forwards them along on the backend requests (and all the authorization logic happens within the relevant service).

Since Gambit doesn't support authorizing requests via OAuth tokens, we'd have to add support to GraphQL to parse those tokens & use that to determine whether or not to use the static Gambit API key for the backend request.

003-graphql.md
DFurnes: Oh man, yeah the playground & its built-in documentation is one of my absolute favorite GraphQL features! Definitely worth a shout-out in here.

003-graphql.md
```diff
@@ -0,0 +1,42 @@
...
+There were a couple of problems we looked to GraphQL to help solve:
+
+ - When we split our old monolith into multiple applications, we introduced conceptual overhead around service boundaries: What resources live in which application? How do things in separate apps relate to each other?
+ - We reimplemented cross-application includes (such as `?include=user` on a list of posts in Rogue) in multiple apps. This caused different backend services to have to act as a "gateway" depending on where a request started.
+ - We found that we would eventually implement the same batching & caching logic in each application (for example, fetching campaigns [in Rogue](https://git.io/fjjGZ) or users [in our now-defunct competitions app](https://github.com/DoSomething/gladiator/blob/534ca9834a5369ededaed56deff33404bbc94e99/app/Repositories/DatabaseUserRepository.php#L62-L163)). In each case, we'd have to reconsider the right order-of-operations and maximum batch size for good performance.
+ - Because each application maintained it's own caches, it was hard to maintain a consistent cache policy. (For example, one application might cache users for an hour while another didn't cache them at all).
```
katiecrane: I'm am SO SORRY but this should be "its" 😬

katiecrane (approved):
📈A well-written and strong argument for why would should use GraphQL!

003-graphql.md
DFurnes: Gah, good catch @katiecrane! (Full disclosure: I think "its vs it's" may be my most frequent Google search and it never sticks. I just can't learn!)

lindsayrmaher (approved):
I whole heartedly approve of this! 📈 ❤️

003-graphql.md
```diff
@@ -0,0 +1,42 @@
...
+
+We should continue to roll out GraphQL and reduce our reliance on custom data-fetching logic!
+
+## Drawbacks/Alternatives
+
+We've traditionally tried to [stick with "boring" technology](http://boringtechnology.club/), because we can follow established best practices & feel confident that we won't likely hit new edge cases. While GraphQL is used at scale [by many of our peers](https://www.apollographql.com/docs/intro/benefits/#ready-for-production), we know that we'll find more surprises than we might with REST & Redux.
+
+There's also an onboarding cost to using GraphQL. While it simplifies a lot of problems we've run into in the past, it also adds to the long list of technologies you need to be familiar with to be a productive DoSomething.org engineer.
```
lindsayrmaher, line 32: While I think there is a cost to onboarding to yet another technology, I think that graphql has enough baseline documentation, tutorials etc that make it fairly enjoyable to learn. (at least in my exp!)

weerd (approved):
Huzzah! 🎉

PR #4, opened 2019-09-11
Add an RFC for our testing methodology.

DFurnes:
This pull request introduces the ["Testing" RFC](https://github.com/DoSomething/rfcs/blob/testing/004-testing.md):

> We write automated tests to help us catch mistakes before they impact our users. Automated tests also allow us to test more thoroughly and quickly than if we were testing "by hand", which lets us iterate faster & more confidently!

This is adapted from the [Automated Testing](https://app.gitbook.com/@dosomething/s/phoenix-documentation/~/drafts/-Lij06iEZbvwsTirohc4/primary/development/automated-testing) documentation I wrote for Phoenix (and GitBooks subsequently deleted). 🤷‍♂ I adjusted some parts to try to generalize it for how we think about testing in all of our apps.

# Open Questions
Does this line up with how others think of testing? What questions are you left with?

### DIFF
--- a/004-testing.md
+++ b/004-testing.md
@@ -2,7 +2,7 @@
 
 We write automated tests to help us catch mistakes before they impact our users. Automated tests also allow us to test more thoroughly and quickly than if we were testing "by hand", which lets us iterate faster & more confidently!
 
-**Discussion:** …
+**Discussion:** https://github.com/DoSomething/rfcs/pull/4
 
 ## Problem
 
--- a/000-testing.md
+++ b/000-testing.md
@@ -1,0 +1,50 @@
+# Testing
+
+We write automated tests to help us catch mistakes before they impact our users. Automated tests also allow us to test more thoroughly and quickly than if we were testing "by hand", which lets us iterate faster & more confidently!
+
+**Discussion:** …
+
+## Problem
+
+We've found that when we skip writing tests, we're less confident in changes we make. That eventually slows down feature development and discourages refactoring. On the other hand, we don't want to spend all of our time writing tests, since we recognize that they're just a tool (and don't directly provide value to our users).
+
+## Proposal
+
+Well, we should write tests, of course! Here's our current "best practices" for testing:
+
+### Philosophy
+There are lots of opinions on how to write tests! We try be pragmatic.
+We write tests to save ourselves from doing tedious QA work, and our tests reflect that: they're the steps that an engineer or PM would use to test a given function, component, API endpoint, or user flow... but translated into a set of instructions so a machine can do it for us!
+
+We [don't practice strict TDD](https://dhh.dk/2014/tdd-is-dead-long-live-testing.html) or pursue 100% test coverage. Instead, we try to get the most "bang for our buck" by writing [fewer higher-level integration tests](https://youtu.be/Fha2bVoC8SE) that test the real functionality that our application exposes, rather than isolated pieces.
+
+In [Phoenix](https://github.com/DoSomething/phoenix), this means the majority of our test coverage comes from [Cypress](https://www.cypress.io/) browser tests, which step through full application flows just like a real user. We also maintain some unit tests, written with [Jest](https://jestjs.io/) & [Enzyme](https://airbnb.io/enzyme/), for more complicated components. In backend Laravel applications, like [Northstar](https://github.com/DoSomething/northstar) and [Rogue](https://github.com/DoSomething/rogue), the majority of our test coverage comes from [PHPUnit](https://laravel.com/docs/5.5/http-tests) functional tests, which simulate real API requests.
+
+### Test Structure
+
+In general, all of our tests follow the same basic structure: "given this scenario, when I do X, I see Y as a result". For example,  ["given this JSON payload, the ContentfulEntry renders the appropriate component"](https://github.com/DoSomething/phoenix-next/blob/dc9713b568060a3a459c715ffb97854321d4ae10/resources/assets/components/ContentfulEntry/ContentfulEntry.test.js#L19-L26) or ["given a user isn't signed up for the campaign, if they click on the Join Now button, then they should see the affirmation modal"](https://github.com/DoSomething/phoenix-next/blob/dc9713b568060a3a459c715ffb97854321d4ae10/cypress/integration/campaign-signup.js#L35-L53).
+
+```js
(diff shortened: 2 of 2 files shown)


DFurnes:
For folks who missed it, it's also probably worth skimming [this proto-RFC from when we started using Cypress in Phoenix](https://github.com/DoSomething/phoenix-next/issues/1437) for context on why we chose that particular tool (and why we're more interested in high-level browser tests than comprehensive component-level unit tests).

There's also still opportunity for some more RFCs for improvements to our current processes,  such as swapping Enzyme to the less fiddly-looking [`react-testing-library`](https://github.com/testing-library/react-testing-library) and making more use of Apollo's [testing helpers](https://www.apollographql.com/docs/react/recipes/testing).

004-testing.md
```diff
@@ -0,0 +1,50 @@
...
+
+**Discussion:** https://github.com/DoSomething/rfcs/pull/4
+
+## Problem
+
+We've found that when we skip writing tests, we're less confident in changes we make. That eventually slows down feature development and discourages refactoring. On the other hand, we don't want to spend all of our time writing tests, since we recognize that they're just a tool (and don't directly provide value to our users).
+
+## Proposal
+
+Well, we should write tests, of course! Here's our current "best practices" for testing:
```
mendelB, line 13: > Well, we should write tests, of course!

mendelB (approved):
👍 Heck. Yes. Thanks for writing this up -- chrystal clear!

> such as swapping Enzyme to the less fiddly-looking react-testing-library

Yeah! I've been reading about this a touch, and it does seem like it'd be a nice choice!

004-testing.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Philosophy
+There are lots of opinions on how to write tests! We try be pragmatic.
+We write tests to save ourselves from doing tedious QA work, and our tests reflect that: they're the steps that an engineer or PM would use to test a given function, component, API endpoint, or user flow... but translated into a set of instructions so a machine can do it for us!
+
+We [don't practice strict TDD](https://dhh.dk/2014/tdd-is-dead-long-live-testing.html) or pursue 100% test coverage. Instead, we try to get the most "bang for our buck" by writing [fewer higher-level integration tests](https://youtu.be/Fha2bVoC8SE) that test the real functionality that our application exposes, rather than isolated pieces.
```
lindsayrmaher, line 19: I definitely think it's difficult to find the right balance for writing tests, but setting some kind of concrete goal would be helpful for us to measure our success. Or even just to see whether 'x' amount of testing is impactful, etc.

004-testing.md
```diff
@@ -0,0 +1,50 @@
...
+   it('should do the things', () => {
+      // 1. Given some state, such as from a mock or factory...
+      // 2. When something happens (click, HTTP request, function call...)
+      // 3. Assert the result matches our expectations!
+   }
+});
+```
+
+### Process
+We like to write tests while adding features or fixing bugs, because writing tests help us "check our work" & demonstrate functionality in pull requests, and because otherwise it's easy to forget! We're still catching up on tests we skipped writing when we originally built Phoenix, Chompy, and other applications.
```
lindsayrmaher, line 38: It might be nice for Vanessa and I to try and bring some existing applications up to speed with testing in order to get more familiar with them! Depending on the focus of Q4, that might be a nice way for us to ramp up.

lindsayrmaher (approved):
Overall I am 100% for more testing and clear goals for our amount of testing on PRs etc etc etc. Thanks for writing this up! 💃

004-testing.md
DFurnes: Yesss! I think this is a wonderful idea. 😸

aaronschachter (approved):
I do wonder a little bit about unit tests. We do have quite a few in Gambit, although for sure we need tons of more integration tests, but they're trickier to write. I think there's a benefit in adding unit tests for any helper functions added if we don't have the bandwidth to add integration tests for a new feature.

004-testing.md
```diff
@@ -0,0 +1,50 @@
...
+
+We try not to merge in code without tests, and never merge in a pull request that introduces a test failure.
+
+### Mocking
+
+Since each of our applications are only small piece of our overall system, we use [mocks](https://en.wikipedia.org/wiki/Mock_object) to avoid relying on external dependencies in our tests. This helps to save us from "flakiness" (such as network issues) and allows us to easily simulate how applications would react to hypothetical scenarios (such as quickly testing how Phoenix's campaign landing page looks for a user with or without an existing campaign signup in Rogue).
```
weerd, line 44: _Side note:_ Not 100% sure if some of our tests account for this or not, but maybe actually accounting for network issues in tests would be useful too.

For example, if Phoenix can't connect to Northstar for whatever reason, does it just stop in its tracks or is there some better user experience in place and maybe some tests that mimic this and check that the expected "fail" scenario is presented. 🤔

mirie (approved):

weerd (approved):
🎉 🚥

004-testing.md
DFurnes: Yeah, I don't think that we have any test coverage for this right now, but it's absolutely worth testing. Adding more tests like this could also go hand in hand with restarting fire drills (e.g. destroy Northstar QA offline one day & see how hard it is to recover), which is something we're hoping to do sometime this year!

VCGibbs:
I like the idea of simplifying and unifying the testing process. I think we're on a good track to prevent future bugs in the future by requiring tests especially for more complex features, front/backend logic. You definitely have my approval!

VCGibbs (approved):

004-testing.md
DFurnes: Yeah, that's totally reasonable. My concern with making test coverage into a tracked metric is that it'd be easy for that to then become a goal in and of itself (when it's really just a tool for helping us to work more efficiently).

At least for now, I like the plan we settled on in our testing sync: to be more thoughtful about making time for writing tests when estimating tickets & to let the "did you add tests" checklist on PRs shame us if we don't. 😼

DFurnes:
Woohoo, six approvals! That certainly looks like consensus to me! 🎉 

I'm going to merge this in, but definitely feel free to drop more comments in here and we can always make amendments in a follow-up PR. Thanks everyone for taking the time to review this and provide thoughtful feedback!

PR #5, opened 2019-09-17
Add a RFC for TailwindCSS Framework

weerd:
This pull request introduces the ["Tailwind CSS Framework" RFC](https://github.com/DoSomething/rfcs/blob/tailwindcss-framework/005-tailwindcss-framework.md):

> We should implement the use of [Tailwind CSS Framework](https://tailwindcss.com/) on our user and admin interfaces to help simplify front-end development across our team.

### DIFF
--- a/005-tailwindcss-framework.md
+++ b/005-tailwindcss-framework.md
@@ -8,15 +8,15 @@ We should implement the use of [Tailwind CSS Framework](https://tailwindcss.com/
 
 It has become more difficult over time for our tech team to confidently update the styling and design patterns of our front-end user interfaces.
 
-While our pattern library, [Forge](https://github.com/DoSomething/forge), worked great for building interfaces with the established patterns it provided, the workflow required (NPM link to test changes locally, publishing releases, updating apps to use new release) did not make updates and rapid iteration to designs patterns feasible, causing Forge to be slow with regular updates and improvement.
+Our pattern library, [Forge](https://github.com/DoSomething/forge), has worked great for building interfaces with the established patterns it provided. However, the workflow required (using NPM link to test changes locally, publishing releases, updating apps to use new release) did not make updates and rapid iteration to designs patterns feasible, causing Forge to be slow with regular updates and improvements. As a result, developers on the team would often end up adding new patterns directly to an application and never migrating them to Forge for use in other applications.
 
 We do appreciate having a library approach to our front-end to help with establishing a system and a shared baseline of patterns, but in practice a CSS library that has simpler updates or just requires less updates while still supporting changing design needs would be an improvement.
 
-As our team structure and focus has changed from dedicated back-end and front-end developers to full-stack developers, we have less team members with extensive, in-depth CSS knowledge, experience or interest in writing cross-browser, responsive style rules. However, our full-stack team still needs to tackle development of features that include full styling of interfaces and showcases the need for a styling framework that is flexible, customizable and empowers our developers to confidently apply styles without the need to code custom style declarations themselves.
+Additionally, as our team structure and focus has changed from dedicated back-end and front-end developers to full-stack developers, we have less team members with extensive, in-depth CSS knowledge, experience or interest in writing cross-browser, responsive style rules. However, our full-stack team still needs to tackle development of features that include full styling of interfaces and showcases the need for a styling framework that is flexible, customizable and empowers our developers to confidently apply styles without the need to code custom style declarations themselves.
 
 ## Proposal
 
-A Utility CSS Framework is a system that provides low-level utility classes that typically map to a corresponding single style rule. They enable you to build custom interfaces entirely with these classes in the HTML for an interface without the need to context-switch languages back-and-forth between HTML and CSS.
+A Utility CSS Framework is a system that provides low-level utility classes that typically map to a corresponding single style rule. They enable you to build custom interfaces entirely with these quick and easy to write classes in the HTML for an interface, without the need to context-switch languages back-and-forth between HTML and CSS.
 
 Additionally, by using utility classes to style HTML, it has the added benefit of abstracting away specificity issues and eliminating problems with the global namespace that can cause style conflicts.
 
@@ -24,26 +24,89 @@ By utilizing a Utility CSS Framework, and more specifically the TailwindCSS Fram
 
 A utility class framework like Tailwind CSS offers the following benefits:
 
-- Establish an API of class names for DoSomething's design style guide.
+- Establishes an API of class names for DoSomething's design style guide.
 - Easily style components and pages by applying a pre-configured library of classes derived from our established style guide directly to HTML.
 - Simplify styling components by significantly reducing the need to write custom CSS, and no more worrying about coming up with unique [BEM-style class names](http://getbem.com/introduction/) as designs change or new designs are developed.
+- Utility classes use functional naming (describe what they do) rather than semantic naming for composed classes, making it easier for full-stack developers to understand what styles are being applied directly to the HTML rather than having to reference a stylesheet and see what series of styles a semantic class is applying.
 - Limits the growth of CSS styles once our collection of utility classes is established and using a library like [PurgeCSS](https://www.purgecss.com/) reduces CSS size further by removing unused utility class definitions in an application.
-- Utility CSS frameworks can not realistic cover every possible style need with their establish library of classes, and Tailwind has useful [plugin system](https://tailwindcss.com/docs/plugins/) to add additional classes needed by a project with full support for outputting classes for responsive support, pseudo-states, etc.
-- Helps keeps CSS scoped to a component, resulting in less worries that a style change could affect styles globally or items along the CSS cascade.
+- Utility CSS frameworks can not realisticly cover every possible style need with their establish library of classes, and Tailwind has a useful [plugin system](https://tailwindcss.com/docs/plugins/) to add additional classes needed by a project with full support for outputting classes for responsive styles, pseudo-states, etc.
+- Helps keep CSS scoped to a component, resulting in less worries that a style change could affect styles globally or items along the CSS cascade.
 - Build more consistent interfaces thanks to utility classes generated based on a configuration file that has been set up using values established in conjunction with the Product Team.
-- Ideally mitigate the use of "magic pixel value numbers" when styling an interface. While such styling may appear, with an established utility class framework in place it _will_ make it very obvious when a developer resorts to using specific pixel value style definitions _instead_ of using an established utility class and can be discussed in a pull request. It will also help in highlighting if we need to potentially add new utility classes to un-magicify aforementioned magic pixel style rules that may appear.
+- Ideally mitigate the use of "magic pixel value numbers" when styling an interface. While such styling may appear, with an established utility class framework in place it _will_ make it very obvious when a developer resorts to using specific pixel value style definitions _instead_ of using established utility classes and can be discussed in a pull request. It will also help in highlighting if we need to potentially add new utility classes to un-magicify aforementioned magic pixel style rules that may appear.
 - Support for [responsiveness](https://tailwindcss.com/docs/responsive-design) and [pseudo-classes](https://tailwindcss.com/docs/pseudo-class-variants) within the library of generated utility classes.
-- Tailwind also provides the ability to extract small components of repeating utility class combinations when needed. While Tailwind encourages always using utility classes in your HTML, there can be scenarios where you may want to extract a component that includes multiple utility classes into a class for use in various places. Since we use React and build most of our interfaces as components, this will be unlikely, but it's a nice feature that Tailwind provides to help with maintainability if some block of HTML is not componentized. See the [Tailwind documentation](https://tailwindcss.com/docs/extracting-components) on extracting components for more information.
+- Tailwind also provides the ability to extract small components of repeating utility class combinations when needed. While Tailwind encourages always using utility classes in your HTML, there can be scenarios where you may want to extract a component that includes multiple utility classes into a class for use in various places. Since we use React and build most of our interfaces as components, this will be unlikely, but it's a nice feature that Tailwind provides to help with maintainability if some blocks of HTML are not componentized. See the [Tailwind documentation](https://tailwindcss.com/docs/extracting-components) on extracting components for more information.
+- Repeating class names in HTML is actually pretty great for GZIP, resulting in smaller HTML files with an already reduced-size CSS file of utility classes.
 
-## Drawbacks/Alternatives
+Additional benefits:
 
-_more to come..._
+- [TailwindCSS](https://github.com/tailwindcss/tailwindcss) is open source so we could contribute back and help further promote open source software!
+- Actively maintained.
+- Available Tailwind CSS IntelliSense plugin for [VS Code](https://marketplace.visualstudio.com/items?itemName=bradlc.vscode-tailwindcss).
+
+## Drawbacks
+
+Once Tailwind CSS is implemented, it will take our full-stack developers a little bit of time to familiarize themselves with the utility class syntax. However the Tailwind CSS documentation is very thorough and easy to search, acting as a great reference. The VS Code extension is also great for suggesting class names based on a project's configuration.
+
+Depending on the complexity of a component, seeing a series of styles displayed as class names within the HTML could become a little harder to read and potentially difficult to spot typos in a class names or potential duplicates. This could be mitigated by establishing a rule whereby class names are added alphabetically or there may be an ESLint rule we could apply; something we can figure out best practices as a team.
+
+There may also be some time to adjust and learn how to best apply utility classes. For example, whether it is best to apply a font-size utility class to a parent container like a `<div>` or to apply it to each individual child `<p>` tag with in parent container. These situations will be tackled and addressed as we encounter them, establishing best practices as a team.
+
+## Alternatives
+
+Tailwind CSS is the recommended utility framework in this proposal, but it is by no means the only utility css framework available. We reviewed and compared a few other alternatives when deciding on which framework to move forward with.
+
+The following list specifies where these alternatives fall short in comparison to Tailwind CSS:
+
+**[Basscss](https://basscss.com/)**
+
+- Less customizable than Tailwind.
+- Not as actively maintained.
+- Extension via NPM pacakage add-ons.
+- Minimal resources and documentation.
+
+**[Beard](http://buildwithbeard.com/)**
+
+- Less customizable than Tailwind.
+- Utilizes SCSS file-based customizations and mixin functions; we have discussed potentially moving away from using SASS in the future.
+- Repository no longer maintained and has been archived.
+
+**[Shed](https://tedconf.github.io/shed-css/index.html)**
+
+- Less customizable than Tailwind.
+- Customized via CSS variables.
+- Not as actively maintained.
+- Shortened class naming convention makes it difficult to immediately understand what the utility class is doing.
+
+**[Tachyons](http://tachyons.io/)**
+
+- Customizing requires installing numerous NPM packages and using command to copy styles into specific project and editing those files.
+- Less customizable than Tailwind.
+- Not as actively maintained.
+- Documentation could be more helpful; (example, the docs are not clear on responsive styles and how to customize).
+
+**[Turretcss](https://turretcss.com/)**
+
+- Contains utility classes but also a framework with composed classes.
+- Customized via CSS variables.
+- Not as actively maintained.
+
+Part of the added appeal of Tailwind CSS is that it uses the power of JavaScript to compose the library of utility classes and has been developed in a way that makes it very extensible, allowing you to write your own custom utilities as plugins, that output classes with all the responsive and pseudo class varieties required.
+
+### Why not CSS-in-JS or CSS Modules?
+
+Using a utility framework like Tailwind CSS does not completely discount the idea of potentially also using an approach with CSS-in-JS like [Emotion](https://emotion.sh/docs/introduction). The basic premise is that you end up writing CSS Styles within the JS component and they are compiled into a custom class specific for that component, helping avoid styling issues with global scope.
+
+CSS Modules is similar in that it helps eliminate issues with the global scope but allows you to write composed classes that you can then import into your JS component and are then compiled into a custom class specific for that component.
+
+However, utility classes inherently avoid styling issues with the global scope and a big part of the aim of using a utility framework is to simplify and minimize the need to write custom CSS in our components; writing custom CSS can potentially result in design variations in the outputted components. Having our full-stack developers select from a predefined list of utility classes helps maintain and enforce consistency across our projects, by limiting styles to those classes available that were generated by rules from a design guide. Rather than having access to millions of colors and a slew of different font-sizes, etc, developers will only have access to the colors and sizes defined by the theme.
+
+We could potentially still see the opportunity to use CSS-in-JS in scenarios where a complex design could require additional styling that utility classes could not address. Some designs where there are very specific pixel size adjustments, etc and it could actually be helpful to maintain those specific cases as custom CSS within the JS for the component. This could establish a process where any required custom CSS is only added specifically to the component where it needs to apply.
 
 ## Forge 7.0
 
 We think that for the next version of Forge, it could embrace the utility class approach and still be the go-to package for styling front-end interfaces on DoSomething applications.
 
-For version 7.0 we propose eliminating all the current patterns in Forge (as our projects switch to using Tailwind, they will stop using these patterns as is and thus no longer needed). Instead Forge would house the main TailwindCSS configuration and provide the generated styles from this custom configuration. Projects that use Forge 7.0 package can include these generated TailwindCSS styles and have access to all the utility classes they otherwise need.
+For version 7.0 we propose eliminating all the current patterns in Forge (as our projects switch to using Tailwind, patterns will switch to being specifically components built out of utility classes). Instead Forge would house the main TailwindCSS configuration and provide the generated styles from this custom configuration. Projects that use Forge 7.0 package can include these generated TailwindCSS styles and have access to all the utility classes they otherwise need.
 
 Additionally, we also propose that Forge could house generic and very common React components that we conclude are useful to use across our DoSomething properties so that there is less duplicated components on different apps.
 
@@ -57,6 +120,7 @@ Additionally, we also propose that Forge could house generic and very common Rea
 
 - [Basscss](https://basscss.com/)
 - [Beard](http://buildwithbeard.com/)
+- [Shed](https://tedconf.github.io/shed-css/index.html)
 - [TailwindCSS](https://tailwindcss.com/docs/utility-first/)
 - [Tachyons](http://tachyons.io/)
 - [Turretcss](https://turretcss.com/)
--- a/005-tailwindcss-framework.md
+++ b/005-tailwindcss-framework.md
@@ -16,18 +16,24 @@ As our team structure and focus has changed from dedicated back-end and front-en
 
 ## Proposal
 
-By utilizing the Tailwind CSS Framework the tech team can work more confidently when building the front-end for different interfaces.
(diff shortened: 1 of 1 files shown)


005-tailwindcss-framework.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Problem
+
+It has become more difficult over time for our tech team to confidently update the styling and design patterns of our front-end user interfaces.
+
+While our pattern library, [Forge](https://github.com/DoSomething/forge), worked great for building interfaces with the established patterns it provided, it has not been updated in some time and with our rapidly changing design needs and team structure, it no longer seems like the ideal solution to continue evolving.
+
+As our team of full-stack developers shifts to doing more work on our projects that includes both back-end and front-end changes, we need a styling framework in place that is flexible, customizable and empowers our developers to confidently use without the need to code custom style declarations themselves.
+
+Additionally, the patterns in Forge are much more specific to how the DoSomething.org website looked in prior iterations. As the design evolved, it has forced us to override many of Forge's rules to be able to replicate changing design needs.
```
DFurnes: This seems like roughly the same argument as the first paragraph, so can maybe be merged in with it.

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,38 @@
...
+
+**Discussion:** <https://github.com/DoSomething/rfcs/pull/5>
+
+## Problem
+
+It has become more difficult over time for our tech team to confidently update the styling and design patterns of our front-end user interfaces.
+
+While our pattern library, [Forge](https://github.com/DoSomething/forge), worked great for building interfaces with the established patterns it provided, it has not been updated in some time and with our rapidly changing design needs and team structure, it no longer seems like the ideal solution to continue evolving.
+
+As our team of full-stack developers shifts to doing more work on our projects that includes both back-end and front-end changes, we need a styling framework in place that is flexible, customizable and empowers our developers to confidently use without the need to code custom style declarations themselves.
```
DFurnes: I _think_ this is what you're getting at here, but I feel it's worth being explicit if so – is the issue that we're all full-stack developers now, and so folks don't have the in-depth CSS experience that they'd perhaps need to "safely" write CSS? (I'd argue that's true, and I think other folks on the team agree!)

To me, one of the biggest problems utility classes would solve for us is that there's one less language to context-switch into & they abstract away some of the nastier parts of CSS (specificity & global namespaces)

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,38 @@
...
+
+We should implement the use of [Tailwind CSS Framework](https://tailwindcss.com/) on our user and admin interfaces to help simplify front-end development across our team.
+
+**Discussion:** <https://github.com/DoSomething/rfcs/pull/5>
+
+## Problem
+
+It has become more difficult over time for our tech team to confidently update the styling and design patterns of our front-end user interfaces.
+
+While our pattern library, [Forge](https://github.com/DoSomething/forge), worked great for building interfaces with the established patterns it provided, it has not been updated in some time and with our rapidly changing design needs and team structure, it no longer seems like the ideal solution to continue evolving.
```
DFurnes: Yeah, I think one of the biggest pain points with Forge was the "library" dev workflow – having to make changes in Forge (and figure out how to `npm link` to see them in Phoenix or wherever) and then publish a release before including the changes in the app. (And something we've had similar pain with elsewhere, such as [this fix for a Gateway issue in Phoenix](https://github.com/DoSomething/phoenix-next/blob/5a63326ea1158022c8096b63c8520198eca803f8/app/Repositories/PostRepository.php#L46-L61)).

(And that's not to say that making it a library wasn't worthwhile – I think it forced us to take more of a "systems" approach to building interfaces then than we have since. And while we weren't good at extracting patterns later on, it was certainly nice to have that shared "baseline" of patterns in every app we work on!)

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,38 @@
...
+## Proposal
+
+By utilizing the Tailwind CSS Framework the tech team can work more confidently when building the front-end for different interfaces.
+
+A utility class framework like Tailwind CSS offers the following benefits:
+
+- We can easily style items and pages by applying a pre-configured library of classes directly to HTML.
+- We can style custom components without needing to write custom CSS, and without needing to invent specific and unique class names for each component.
+- Once we establish our collection of utility classes, a project's CSS stops growing, since there is less need for custom CSS.
+- Helps keeps CSS scoped to a component, resulting in less worries that a style change could affect styles globally.
```
DFurnes: I'm also very interested in using [CSS Modules](https://github.com/css-modules/css-modules) for this (which isn't necessarily incompatible with Tailwind), since it solves for the global namespace issue by programmatically creating classnames & letting us do dead-code elimination (since there wouldn't be implicit string references to classes).

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,38 @@
...
+- We will be building interfaces with proper constraints defined by the framework's configuration; no more magic pixel value numbers.
```
DFurnes: How would Tailwind help us avoid this? I feel like we had similar tools (like `$base-spacing`) in Forge, and the problem was more about miscommunications where developers would try to match the "pixel perfect" mock and introduce magic numbers to do so (rather than pushing back on the mocks).

DFurnes:
And just posting this as a follow-up to our IRL conversation the other day:

I'd love to see a "Drawbacks/Alternatives" section in this RFC that addresses why we'd choose Tailwind over other CSS utility frameworks (like [Tachyons](https://tachyons.io/), [basscss](https://basscss.com/), or [turretcss](https://turretcss.com/)) and why we'd choose this approach over a CSS-in-JS solution like [Emotion](https://emotion.sh/docs/introduction).

005-tailwindcss-framework.md
mendelB: ➕

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,38 @@
...
+
+As our team of full-stack developers shifts to doing more work on our projects that includes both back-end and front-end changes, we need a styling framework in place that is flexible, customizable and empowers our developers to confidently use without the need to code custom style declarations themselves.
+
+Additionally, the patterns in Forge are much more specific to how the DoSomething.org website looked in prior iterations. As the design evolved, it has forced us to override many of Forge's rules to be able to replicate changing design needs.
+
+## Proposal
+
+By utilizing the Tailwind CSS Framework the tech team can work more confidently when building the front-end for different interfaces.
+
+A utility class framework like Tailwind CSS offers the following benefits:
```
mendelB, line 25: Do you think it'd be a good idea to include a link the ole Adam article for the uninitiated? https://adamwathan.me/css-utility-classes-and-separation-of-concerns/

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,38 @@
...
+
+- We can easily style items and pages by applying a pre-configured library of classes directly to HTML.
+- We can style custom components without needing to write custom CSS, and without needing to invent specific and unique class names for each component.
+- Once we establish our collection of utility classes, a project's CSS stops growing, since there is less need for custom CSS.
+- Helps keeps CSS scoped to a component, resulting in less worries that a style change could affect styles globally.
+- We will be building interfaces with proper constraints defined by the framework's configuration; no more magic pixel value numbers.
+- The utilities would help developers adhere to the predefined design system, keeping our interfaces consistent.
+- Support for responsive design and pseudo-classes within the series of generated utility classes.
+- If needed, we can extract components when it makes sense.
```
mendelB: I feel like this could perhaps use some clarification. The feeling of this proposal seems to me to be that we're sort of doing away with components entirely, with potential pushback being 'but what about e.g. buttons where we do want some centralized styling as opposed to individual utility classes for each button implementation?'. 

Could we perhaps just expand on this idea of 'extracting components' a bit with maybe an example of when this would make sense to do and how that'd work with utilities? 

(This was something I personally was confused about, which you helped clarify a bunch during your CSS overview session which I found very helpful!)

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Forge 7.0
+
+We think that for the next version of Forge, it could embrace the utility class approach and still be the go-to package for styling front-end interfaces on DoSomething applications.
+
+For version 7.0 we propose eliminating all the current patterns in Forge (as our projects switch to using Tailwind, they will stop using these patterns as is and thus no longer needed). Instead Forge would house the main TailwindCSS configuration and provide the generated styles from this custom configuration. Projects that use Forge 7.0 package can include these generated TailwindCSS styles and have access to all the utility classes they otherwise need.
```
mendelB: What would this mean for much of the styling on Phoenix where we rely on current Forge patterns -- would we just need to do a clean sweep of those classes in favor of the Tailwind classes before updating to `7.0`? (Though how would we do that _without_ having Tailwind installed from 7.0 in the first place?)

005-tailwindcss-framework.md
katiecrane: ➕

005-tailwindcss-framework.md
katiecrane: I think it'd be helpful if that link was included somewhere as well!

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Proposal
+
+By utilizing the Tailwind CSS Framework the tech team can work more confidently when building the front-end for different interfaces.
+
+A utility class framework like Tailwind CSS offers the following benefits:
+
+- We can easily style items and pages by applying a pre-configured library of classes directly to HTML.
+- We can style custom components without needing to write custom CSS, and without needing to invent specific and unique class names for each component.
+- Once we establish our collection of utility classes, a project's CSS stops growing, since there is less need for custom CSS.
```
katiecrane: This bit confuses me because the first point sounds like we don't need to write custom CSS and the second point sounds like we do. Could you elaborate on why having to write some custom CSS is still helpful? I don't think I've quite wrapped my head around it.

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,38 @@
...
+- Support for responsive design and pseudo-classes within the series of generated utility classes.
+- If needed, we can extract components when it makes sense.
+
+## Forge 7.0
+
+We think that for the next version of Forge, it could embrace the utility class approach and still be the go-to package for styling front-end interfaces on DoSomething applications.
+
+For version 7.0 we propose eliminating all the current patterns in Forge (as our projects switch to using Tailwind, they will stop using these patterns as is and thus no longer needed). Instead Forge would house the main TailwindCSS configuration and provide the generated styles from this custom configuration. Projects that use Forge 7.0 package can include these generated TailwindCSS styles and have access to all the utility classes they otherwise need.
+
+Additionally, we also propose that Forge could house generic and very common React components that we conclude are useful to use across our DoSomething properties so that there is less duplicated components on different apps.
```
katiecrane, line 113: I see the `<Query>` component being a good one for this!

005-tailwindcss-framework.md
weerd: I think the process would be:

1. Implement Tailwind in Phoenix.
2. Proceed with updating Phoenix to utilize Tailwind classes instead of Forge or custom ones (so Forge no longer plays a role).
3. Update Forge to 7.0 which will consist of the Tailwind config (pulled from Phoenix) and any components deemed reusable across platforms.
4. Re-implement Forge (v7.0) into Phoenix and remove the Tailwind config and reusable components from Phoenix since now Forge v7.0 will provide them.

I _think_ that's the general gist.

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,126 @@
...
+
+A utility class framework like Tailwind CSS offers the following benefits:
+
+- Establishes an API of class names for DoSomething's design style guide.
+- Easily style components and pages by applying a pre-configured library of classes derived from our established style guide directly to HTML.
+- Simplify styling components by significantly reducing the need to write custom CSS, and no more worrying about coming up with unique [BEM-style class names](http://getbem.com/introduction/) as designs change or new designs are developed.
+- Utility classes use functional naming (describe what they do) rather than semantic naming for composed classes, making it easier for full-stack developers to understand what styles are being applied directly to the HTML rather than having to reference a stylesheet and see what series of styles a semantic class is applying.
+- Limits the growth of CSS styles once our collection of utility classes is established and using a library like [PurgeCSS](https://www.purgecss.com/) reduces CSS size further by removing unused utility class definitions in an application.
```
katiecrane, line 31: > Limits the growth of CSS styles once our collection of utility classes is established

Could you expand on this piece? Does using a Utility CSS Framework just amount to writing less CSS in general?

weerd: Yep! This might be easier to explain with an example:

```css
.callout  {
    background-color: white;
    border: 1px solid gray;
    border-radius: 4px;
    display: block;
    font-size: 22px;
}

.card  {
    background-color: white;
    border: 1px solid gray;
    border-radius: 4px;
    display: block;
    font-size: 18px;
}
```

Some of the above styles could get repeated multiple times if we add another block-like composed class like `.media`, which may be slightly different but have a few things that overlap. We could try and break each down farther for some sort of very generic class that has just the `background-color`, `display` and `border` definitions, but then you'd have the CSS that builds those components out defined in different files. Makes it much harder to maintain and harder to keep track of if we don't have a dedicated front-end team. Some of these effects we see now in our current approaches.

With a utility css framework, the classes you'd have for the above would be the following and they would only really need to get defined _once_:

```css
.bg-white {
    background-color: white;
}

.border {
    border-width: 1px;
}

.border-gray-200 {
    border-color: #edf2f7;
}

.border-solid {
    border-style: solid;
}

.rounded {
    border-radius: .25rem;
}

.block {
    display: block;
}

.text-base {
    font-size: 1rem;
}

.text-lg {
    font-size: 1.125rem;
} 
```

In contrast, I did a quick search on Phoenix and `display: block;` is used 31 times (kind of surprised it's only that, likely because we started using some utility classes). Once tailwind is implemented, it would ideally only be defined once and then the class that defines it `.block`, would be used within the HTML only.

So once we have our set of utility classes established, our CSS shouldn't really grow much over time since we'll be building by just reusing class names to define styles.

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,126 @@
...
+
+We should implement the use of [Tailwind CSS Framework](https://tailwindcss.com/) on our user and admin interfaces to help simplify front-end development across our team.
+
+**Discussion:** <https://github.com/DoSomething/rfcs/pull/5>
+
+## Problem
+
+It has become more difficult over time for our tech team to confidently update the styling and design patterns of our front-end user interfaces.
+
+Our pattern library, [Forge](https://github.com/DoSomething/forge), has worked great for building interfaces with the established patterns it provided. However, the workflow required (using NPM link to test changes locally, publishing releases, updating apps to use new release) did not make updates and rapid iteration to designs patterns feasible, causing Forge to be slow with regular updates and improvements. As a result, developers on the team would often end up adding new patterns directly to an application and never migrating them to Forge for use in other applications.
```
mendelB, line 11: Ya know, I feel like this is a pain point that won't be completely solved by switching to Tailwind. Of course, by using utility-based classes over Forge patterns the inherent dynamic dependability between the specific app and Forge will be lessened, causing this problem to be lessened (_way_ lessened!) by definition, but in terms of updating and adding newly defined customizations (both formal pixel/font-size/color definitions & overrides, and extracted common components & updates) in Forge 7.0 -- it'll probably _still_ feel a lot more convenient to just implement it within the respective app itself vs the whole workflow required to update Forge! 

I hope I'm making sense. And def not an objection to this, but more so calling out that this may be something to bear in mind and we forge (😏) ahead with Tailwind. It's obviously not the magic pill for _all_ our problems 😥 but it's pretty damn close 🙏

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,126 @@
...
+
+By utilizing a Utility CSS Framework, and more specifically the TailwindCSS Framework, the tech team can work more confidently when building the front-end for different interfaces across our platforms.
+
+A utility class framework like Tailwind CSS offers the following benefits:
+
+- Establishes an API of class names for DoSomething's design style guide.
+- Easily style components and pages by applying a pre-configured library of classes derived from our established style guide directly to HTML.
+- Simplify styling components by significantly reducing the need to write custom CSS, and no more worrying about coming up with unique [BEM-style class names](http://getbem.com/introduction/) as designs change or new designs are developed.
+- Utility classes use functional naming (describe what they do) rather than semantic naming for composed classes, making it easier for full-stack developers to understand what styles are being applied directly to the HTML rather than having to reference a stylesheet and see what series of styles a semantic class is applying.
```
mendelB, line 30: I also love how this functional naming (and the process of looking up & learning the proper Tailwind class names for respective stylings) seems to me to also help enhance the CSS knowledge thereof! (I feel like I've learned or enhanced my understanding of certain styling when e.g. I've had to research and decide which Tailwind class to use & think about what it was doing vs the other similar class -- thus thinking more carefully about the styling functionality itself! While with raw CSS, I might have just played around with some padding or whatever and just mindlessly applied a rule)

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,126 @@
...
+- Support for [responsiveness](https://tailwindcss.com/docs/responsive-design) and [pseudo-classes](https://tailwindcss.com/docs/pseudo-class-variants) within the library of generated utility classes.
+- Tailwind also provides the ability to extract small components of repeating utility class combinations when needed. While Tailwind encourages always using utility classes in your HTML, there can be scenarios where you may want to extract a component that includes multiple utility classes into a class for use in various places. Since we use React and build most of our interfaces as components, this will be unlikely, but it's a nice feature that Tailwind provides to help with maintainability if some blocks of HTML are not componentized. See the [Tailwind documentation](https://tailwindcss.com/docs/extracting-components) on extracting components for more information.
```
mendelB, line 37: I'm trying to wrap my head -generally- around the idea of us replacing some of these entrenched and often used Forge patterns with Tailwind & components.

So by 
> there can be scenarios where you may want to extract a component that includes multiple utility classes into a class for use in various places. Since we use React and build most of our interfaces as components, this will be unlikely

Would this mean something like, instead of e.g. the Forge [`.text-field`](https://github.com/DoSomething/forge/blob/dev/scss/_components/_forms/_text-field.scss) that class we use [quite](https://github.com/DoSomething/phoenix-next/blob/f4e762488824085da695e74ee314001212f12d31/resources/assets/components/actions/PhotoSubmissionAction/PhotoSubmissionAction.js#L332) [often](https://github.com/DoSomething/phoenix-next/blob/f4e762488824085da695e74ee314001212f12d31/resources/assets/components/actions/TextSubmissionAction/TextSubmissionAction.js#L185) around Phoenix. We might institute an `Input` React component with the base styling, and then customize with utility classes from wherever it gets used, vs. adding a `text-field` Tailwind component which _would_ be similar to the current Forge pattern?

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,126 @@
...
+- Repeating class names in HTML is actually pretty great for GZIP, resulting in smaller HTML files with an already reduced-size CSS file of utility classes.
+
+Additional benefits:
+
+- [TailwindCSS](https://github.com/tailwindcss/tailwindcss) is open source so we could contribute back and help further promote open source software!
```
mendelB, line 42: 💯

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,126 @@
...
+However, utility classes inherently avoid styling issues with the global scope and a big part of the aim of using a utility framework is to simplify and minimize the need to write custom CSS in our components; writing custom CSS can potentially result in design variations in the outputted components. Having our full-stack developers select from a predefined list of utility classes helps maintain and enforce consistency across our projects, by limiting styles to those classes available that were generated by rules from a design guide. Rather than having access to millions of colors and a slew of different font-sizes, etc, developers will only have access to the colors and sizes defined by the theme.
+
+We could potentially still see the opportunity to use CSS-in-JS in scenarios where a complex design could require additional styling that utility classes could not address. Some designs where there are very specific pixel size adjustments, etc and it could actually be helpful to maintain those specific cases as custom CSS within the JS for the component. This could establish a process where any required custom CSS is only added specifically to the component where it needs to apply.
```
mendelB, line 103: This does sound like a nice complementary pattern!

mendelB (approved):
📰 Great read! I'm super psyched about this and have absolutely loved working with Tailwind so far. Excited to see this vision manifest -- it's been a [long time coming](https://github.com/DoSomething/phoenix-next/pull/693#pullrequestreview-98350999) 😁

005-tailwindcss-framework.md
weerd: Yeah, I'm not 100% sure on what the answer is there but definitely something we'll figure out as we go. 

Personally I think having a React `Input` component would be tooooo granular and could bite us if in some scenario we need a different approach to a textfield 🤷‍♂ There's something nice about seeing a form and the direct HTML that builds most of it and is easily editable within context of other form elements.

So maybe in this case a composed Tailwind class is the way to go, which could mean it also allows flexibility to not use the classname in some scenario and instead build out with utilities.

Definitely a good question and we'll just have to see where the dust settles 💃

005-tailwindcss-framework.md
weerd: > in terms of updating and adding newly defined customizations (both formal pixel/font-size/color definitions & overrides, and extracted common components & updates) in Forge 7.0 -- it'll probably still feel a lot more convenient to just implement it within the respective app itself vs the whole workflow required to update Forge!

True, but in terms of _what_ and _when_ these updates happen I believe they would be much less frequent. It's not often that we would need to add a new color or a new font-size, etc. And same kind of goes for components... we'd only really want to include components in Forge that are generic and whose application could be used in different projects which are also less common in our projects.

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,126 @@
...
+
+Tailwind CSS is the recommended utility framework in this proposal, but it is by no means the only utility css framework available. We reviewed and compared a few other alternatives when deciding on which framework to move forward with.
+
+The following list specifies where these alternatives fall short in comparison to Tailwind CSS:
+
+**[Basscss](https://basscss.com/)**
+
+- Less customizable than Tailwind.
+- Not as actively maintained.
+- Extension via NPM pacakage add-ons.
```
aaronschachter: Nit: mispelled `package`

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,126 @@
...
+
+Depending on the complexity of a component, seeing a series of styles displayed as class names within the HTML could become a little harder to read and potentially difficult to spot typos in a class names or potential duplicates. This could be mitigated by establishing a rule whereby class names are added alphabetically or there may be an ESLint rule we could apply; something we can figure out best practices as a team.
+
+There may also be some time to adjust and learn how to best apply utility classes. For example, whether it is best to apply a font-size utility class to a parent container like a `<div>` or to apply it to each individual child `<p>` tag with in parent container. These situations will be tackled and addressed as we encounter them, establishing best practices as a team.
+
+## Alternatives
+
+Tailwind CSS is the recommended utility framework in this proposal, but it is by no means the only utility css framework available. We reviewed and compared a few other alternatives when deciding on which framework to move forward with.
+
+The following list specifies where these alternatives fall short in comparison to Tailwind CSS:
```
aaronschachter, line 58: I think this will be really helpful to have documented, thanks for listing all of these out.

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,126 @@
...
+
+**[Basscss](https://basscss.com/)**
+
+- Less customizable than Tailwind.
```
aaronschachter, line 62: Totally a nice-to-have, but if you remember off-hand per looking at each - it might be helpful to list a few reasons why for each of these are less customizable (this reason occurs in multiple alternatives)

aaronschachter (approved):
Reads great, sounds great! Awesome work.

005-tailwindcss-framework.md
weerd: it's a 🍰 package, lol 🤦‍♂

005-tailwindcss-framework.md
katiecrane: THANK YOU this example is extremely helpful!! Thank you for taking the time to spell out a real world example, really helped me understand it a lot better!

katiecrane (approved):
I'm on board! Well written and convincing!

005-tailwindcss-framework.md
DFurnes: Yeah, I agree that this is a good step forward but doesn't fully solve that problem. At some point, I'd like to experiment with sharing code between apps [via a monorepo](https://docs.google.com/document/d/1a3e0Wpv9qeRscspdbN4wVR4Chz5oJFb9HH69x2pngSg/edit) so that we don't have to worry about independent versioning for these shared resources.

005-tailwindcss-framework.md
DFurnes: +1, this is a great example!

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,126 @@
...
+
+Additionally, by using utility classes to style HTML, it has the added benefit of abstracting away specificity issues and eliminating problems with the global namespace that can cause style conflicts.
+
+By utilizing a Utility CSS Framework, and more specifically the TailwindCSS Framework, the tech team can work more confidently when building the front-end for different interfaces across our platforms.
+
+A utility class framework like Tailwind CSS offers the following benefits:
+
+- Establishes an API of class names for DoSomething's design style guide.
+- Easily style components and pages by applying a pre-configured library of classes derived from our established style guide directly to HTML.
+- Simplify styling components by significantly reducing the need to write custom CSS, and no more worrying about coming up with unique [BEM-style class names](http://getbem.com/introduction/) as designs change or new designs are developed.
```
DFurnes, line 29: Naming always _is_ the hardest problem! 🙉

005-tailwindcss-framework.md
DFurnes, line 31: PurgeCSS leaves a little to be desired, since it's just naively looking for selectors by string matching over a directory of source files. And so we can't do code-splitting for those styles (which may not matter if we mainly rely on utility classes), or extract critical CSS per page (more interesting, IMHO).

(Not a deal-breaker for Tailwind, but a good incentive to experiment with Emotion like you mentioned below!)

005-tailwindcss-framework.md
```diff
@@ -0,0 +1,126 @@
...
+- Tailwind also provides the ability to extract small components of repeating utility class combinations when needed. While Tailwind encourages always using utility classes in your HTML, there can be scenarios where you may want to extract a component that includes multiple utility classes into a class for use in various places. Since we use React and build most of our interfaces as components, this will be unlikely, but it's a nice feature that Tailwind provides to help with maintainability if some blocks of HTML are not componentized. See the [Tailwind documentation](https://tailwindcss.com/docs/extracting-components) on extracting components for more information.
+- Repeating class names in HTML is actually pretty great for GZIP, resulting in smaller HTML files with an already reduced-size CSS file of utility classes.
+
+Additional benefits:
+
+- [TailwindCSS](https://github.com/tailwindcss/tailwindcss) is open source so we could contribute back and help further promote open source software!
+- Actively maintained.
+- Available Tailwind CSS IntelliSense plugin for [VS Code](https://marketplace.visualstudio.com/items?itemName=bradlc.vscode-tailwindcss).
```
DFurnes, line 44: Sweeeeet! ✏️

DFurnes (approved):
Really excited to start playing around with this! 💅

005-tailwindcss-framework.md
weerd: I think more-so in comparison to the list of Tailwind's capabilities and customizability the others are less customizable. Tailwind's easy customizability via a JS config object, overriding or extending the library of classes it generates, extensibility via plugins, and ability to still create composed classes all give it a lot of versatility compared to the others which have simpler customization options. I can include the above statement in the description though to make it clearer! Thanks for the suggestion @aaronschachter 😄

PR #6, opened 2019-10-31
Add a RFC for GraphQL Pagination.

DFurnes:
This pull request introduces the ["GraphQL Pagination"](https://github.com/DoSomething/rfcs/blob/graphql-pagination/006-graphql-pagination.md) RFC:

> We should use [GraphQL Cursor Connections](https://facebook.github.io/relay/graphql/connections.htm#) to paginate results in GraphQL. 

# Open Questions
- Are there pain points we've run into with pagination that this doesn't solve?
- Is it worth keeping the simple queries (e.g. `posts: [Post]`) or should we deprecate them?

### DIFF
--- a/006-graphql-pagination.md
+++ b/006-graphql-pagination.md
@@ -1,6 +1,6 @@
 # GraphQL Pagination
 
-We should use [Relay Cursor Connections](https://facebook.github.io/relay/graphql/connections.htm#) to paginate results in GraphQL. 
+We should use [GraphQL Cursor Connections](https://facebook.github.io/relay/graphql/connections.htm#) to paginate results in GraphQL. 
 
 **Discussion:** https://github.com/DoSomething/rfcs/pull/6
 
@@ -22,7 +22,7 @@ This gives us no way to know whether there's another page of results without mak
 
 ## Proposal
 
-We should use [Relay Cursor Connections](https://facebook.github.io/relay/graphql/connections.htm#) to paginate results in GraphQL, and add REST API support to paginate via cursors (e.g. "get me all the results after #1015" rather than "get me the tenth page of results") to Northstar and Rogue.
+We should use [GraphQL Cursor Connections](https://facebook.github.io/relay/graphql/connections.htm#) to paginate results in GraphQL, and add REST API support to paginate via cursors (e.g. "get me all the results after #1015" rather than "get me the tenth page of results") to Northstar and Rogue.
 
 ```gql
 {
--- a/006-graphql-pagination.md
+++ b/006-graphql-pagination.md
@@ -1,0 +1,84 @@
+# GraphQL Pagination
+
+We should use [Relay Cursor Connections](https://facebook.github.io/relay/graphql/connections.htm#) to paginate results in GraphQL. 
+
+**Discussion:** https://github.com/DoSomething/rfcs/pull/6
+
+## Problem
+
+We currently return plain lists, like `[Post]`, for paginated GraphQL queries:
+
+```gql
+{
+  posts(count: 10, page: $page) {
+    id
+    type
+    url
+  }
+}
+```
+
+This gives us no way to know whether there's another page of results without making a request to check. Because we built these on top of standard `OFFSET` pagination, there are also some performance issues with larger datasets & the potential for duplicate results to appear (described in more detail under "Alternatives").
+
+## Proposal
+
+We should use [Relay Cursor Connections](https://facebook.github.io/relay/graphql/connections.htm#) to paginate results in GraphQL, and add REST API support to paginate via cursors (e.g. "get me all the results after #1015" rather than "get me the tenth page of results") to Northstar and Rogue.
+
+```gql
+{
+  paginatedPosts(first: 10, after: $cursor) {
+    edges {
+      cursor
+      node {
+        id
+        type
+        url
+      }
+    }
+    pageInfo {
+      hasNextPage
+    }
+  }
+}
+```
+
+This relies on generating "opaque" cursors for each element in the results. The cursor, e.g. `Nzk1MS4yNDA3Nw==` encodes information about where the item appears in the results (the ID & the sorted column value, if applicable).
+
+## Drawbacks
+
+This is certainly more verbose than using standard lists! I could see a good argument for having both types of queries available (say, `campaigns` and `paginatedCampaigns`) so we can use each where it makes the most sense.
+
+Using cursors for pagination also requires some REST API updates (to generate the cursor for each item in a paginated response and then to build the database query based on the given `?cursor` query string).
+
+## Alternatives
+
+There are [multiple approaches to pagination with GraphQL](https://www.apollographql.com/docs/react/data/pagination/).
+
+The simplest option is to return a plain list, e.g. `[Post]`, but this has a big downside – we have no way of knowing when we've reached the end of the result set (until we make a query for the next page and get zero results). This is why campaign galleries always show "Load More" even when they don't have more pages to load:
+
+```gql
+posts(…args): [Post]
+```
+
+We could build a home-grown solution that adds `hasNextPage` and `hasPreviousPage` fields:
+
+```gql
+posts(page: Int, ...): {
+  hasNextPage: Bool!
+  hasPreviousPage: Bool!
+  results: [Post]
+}
+```
+
+This solves our main issue (not knowing when we've reached the end of our result set), but still has some downsides because we're using traditional numbered pages:
+- First, using `?page=N` gets progressively slower on large collections, because the database has to iterate over N pages of results before "slicing" out the new page. We've run into this problem when Quasar scrapes Northstar's `v2/users` endpoint (and responses get slower and slower as it pages through results).
+- Second, we can accidentally skip over data! If we're sorting so that new items are shown first (like our campaign galleries), users will see duplicates if paginating when new results are added. (Since if we offset over the first 10 posts but then a new one is added, we'll see that 10th post again on the second page).
+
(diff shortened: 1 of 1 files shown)


006-graphql-pagination.md
```diff
@@ -0,0 +1,84 @@
...
+  hasPreviousPage: Bool!
+  results: [Post]
+}
+```
+
+This solves our main issue (not knowing when we've reached the end of our result set), but still has some downsides because we're using traditional numbered pages:
+- First, using `?page=N` gets progressively slower on large collections, because the database has to iterate over N pages of results before "slicing" out the new page. We've run into this problem when Quasar scrapes Northstar's `v2/users` endpoint (and responses get slower and slower as it pages through results).
+- Second, we can accidentally skip over data! If we're sorting so that new items are shown first (like our campaign galleries), users will see duplicates if paginating when new results are added. (Since if we offset over the first 10 posts but then a new one is added, we'll see that 10th post again on the second page).
+
+Neither of those are deal-breakers, but using cursors handily solves them for us! We also get the benefit of following a specification that has worked well for other folks, like Facebook and [GitHub](https://developer.github.com/v4/), rather than blazing our own path.
```
mendelB, line 79: So, if I understand correctly, there doesn't seem to be the _most pressing_ of reasons for us to adopt cursor pagination since as you outline -- the two persistent issues seem -at least somewhat- minor (although I feel like I've actually noticed the second one in galleries before) and wouldn't _generally_ affect most of our apps. But nevertheless, it'd be a nice win, and follow a neat formal spec. Am I digesting the argument properly?

006-graphql-pagination.md
```diff
@@ -0,0 +1,84 @@
...
+      }
+    }
+    pageInfo {
+      hasNextPage
+    }
+  }
+}
+```
+
+This relies on generating "opaque" cursors for each element in the results. The cursor, e.g. `Nzk1MS4yNDA3Nw==` encodes information about where the item appears in the results (the ID & the sorted column value, if applicable).
```
mendelB, line 45: Maybe worth linking to where you've already implemented this in graphql & rogue for a concrete example?

006-graphql-pagination.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Drawbacks
+
+This is certainly more verbose than using standard lists! I could see a good argument for having both types of queries available (say, `campaigns` and `paginatedCampaigns`) so we can use each where it makes the most sense.
+
+Using cursors for pagination also requires some REST API updates (to generate the cursor for each item in a paginated response and then to build the database query based on the given `?cursor` query string).
```
mendelB, line 51: Supporting both seems smart, especially since that'd allow us to be flexibly 'backwards compatible' as we migrate to cursors!

006-graphql-pagination.md
```diff
@@ -0,0 +1,84 @@
...
+    pageInfo {
+      hasNextPage
+    }
+  }
+}
+```
+
+This relies on generating "opaque" cursors for each element in the results. The cursor, e.g. `Nzk1MS4yNDA3Nw==` encodes information about where the item appears in the results (the ID & the sorted column value, if applicable).
+
+## Drawbacks
```
mendelB, line 47: I'm not sure if this drawback is worth explicitly calling out, but in the Shopify article they mention:
> Relative cursor pagination remembers where you were so that each request after the first continues from where the previous request left off. The downside is that you can no longer jump to a specific page.

Though I doubt that's a big deal for us in practice?

006-graphql-pagination.md
```diff
@@ -0,0 +1,84 @@
...
+- First, using `?page=N` gets progressively slower on large collections, because the database has to iterate over N pages of results before "slicing" out the new page. We've run into this problem when Quasar scrapes Northstar's `v2/users` endpoint (and responses get slower and slower as it pages through results).
+- Second, we can accidentally skip over data! If we're sorting so that new items are shown first (like our campaign galleries), users will see duplicates if paginating when new results are added. (Since if we offset over the first 10 posts but then a new one is added, we'll see that 10th post again on the second page).
+
+Neither of those are deal-breakers, but using cursors handily solves them for us! We also get the benefit of following a specification that has worked well for other folks, like Facebook and [GitHub](https://developer.github.com/v4/), rather than blazing our own path.
+
+## References
+
+- [GraphQL Documentation: Pagination](https://graphql.org/learn/pagination/)
+- [Shopify Engineering: "Pagination with Relative Cursors"](https://engineering.shopify.com/blogs/engineering/pagination-relative-cursors)
```
mendelB, line 84: I really enjoyed this article and it kinda blew my mind!

006-graphql-pagination.md
DFurnes: Yeah exactly! These are both frustrating (but not particularly high-priority) bugs, but it's nice that this "battle-tested" approach solves them both for us.

006-graphql-pagination.md
DFurnes: Yes! This is implemented on Rogue's campaigns & posts endpoints (since we need to be able to adjust the database query for pagination) in the [`HasCursor` trait](https://github.com/DoSomething/rogue/blob/564255e20b15942d9a0c05d7fa1d169c50e9c60a/app/Models/Traits/HasCursor.php).

For unsorted lists, the cursor is simply the ID (so for example `NA==`, which base-64 decoded is just `4`). So we'd use `cursor[after]=NA==` to get results after ID 4.

For sorted lists, the cursor is the ID & sorted column value (so for example, `OTAyOC41OTA3Nw==` becomes `9028.59077` for campaign ID 9028 with 59,077 pending posts). This allows us to get results after the sorted column value (and handle cases where multiple items have the same sorted column value, like a bunch of campaigns with zero pending posts).

006-graphql-pagination.md
DFurnes: Oh, I definitely think that's worth calling out! I'll add that in here.

DFurnes:
For a look at how this works in practice, you can take a peek at the following PRs:

1. DoSomething/rogue#976, which adds support for paginating & encoding cursors in Rogue's REST `v3/signups` API (by reusing a trait I'd added in DoSomething/rogue#937).
2. DoSomething/graphql#185, which adds a `paginatedSignups` query that uses the cursor and `?after[cursor]=` REST support to provide Relay-style pagination for this resource in GraphQL.
3. DoSomething/rogue#977, which implements the actual paginated user-interface element.

006-graphql-pagination.md
lindsayrmaher: I second keeping both to be backwards compatible and implement this approach over time.

006-graphql-pagination.md
lindsayrmaher: I do think this is a good flag, however I think in the implementations we've had recently we are moving away from the `next` & `previous` page and more towards the infinite scroll, so it could be a non issue.

lindsayrmaher (approved):
I think the biggest drawback as mentioned is removing the ability to jump to certain pages, but since this was drafted i think we've found a clean solution for user experience by using an infinite scroll experience rather than pages. 

In the infinite scroll implementation, we do run the risk of an incredibly long list of results depending on what we're querying (campaigns for example). But as long as we're cognizant that we need to make the ability to scroll to the top easily accessible it shouldn't be a gamechanger. 

I'm very into the Pagination approach laid out here!

006-graphql-pagination.md
mendelB: _So_ helpful, thanks for the link and explanation!

mendelB (approved):
I've appreciated using cursor pagination in practice over [on Phoenix](https://github.com/DoSomething/phoenix-next/blob/61bc54caff1dd688399279563a5248b4e5f8438b/resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js) and think this is a great idea! 👍 

Thanks for an awesome write-up!

006-graphql-pagination.md
```diff
@@ -0,0 +1,84 @@
...
+This gives us no way to know whether there's another page of results without making a request to check. Because we built these on top of standard `OFFSET` pagination, there are also some performance issues with larger datasets & the potential for duplicate results to appear (described in more detail under "Alternatives").
+
+## Proposal
+
+We should use [Relay Cursor Connections](https://facebook.github.io/relay/graphql/connections.htm#) to paginate results in GraphQL, and add REST API support to paginate via cursors (e.g. "get me all the results after #1015" rather than "get me the tenth page of results") to Northstar and Rogue.
+
+```gql
+{
+  paginatedPosts(first: 10, after: $cursor) {
+    edges {
```
katiecrane, line 30: should this be `posts`?

006-graphql-pagination.md
katiecrane: I don't think this update ever got added in?

006-graphql-pagination.md
DFurnes: This is the "standard" `{ edges { node, cursor }, pageInfo }` structure from the [GraphQL Cursor Connection](https://facebook.github.io/relay/graphql/connections.htm) spec. The terminology is a little awkward, but it's nice that it allows us to handle pagination logic the same way no matter what we're paginating!

katiecrane: Thanks! I was looking there and they only have the one example similar to this at the top - in this example, is a `node` a `post`?

DFurnes: Yep! In this example, each `node` would be a `Post`.

katiecrane: Got it! Thank you!!!

006-graphql-pagination.md
DFurnes: Oops! Added in 9cc2dae.

katiecrane (approved):
I dig it!

PR #7, opened 2019-12-11
Add RFC for using an on-call rotation.

DFurnes:
This pull request introduces the ["On-Call" RFC](https://github.com/DoSomething/rfcs/blob/on-call/007-on-call.md):

> We should use [PagerDuty](http://pagerduty.com) to centralize our alerting in one place & manage an on-call rotation.

# Open Questions
 - After reading this, do you feel comfortable being on-call? What would make you feel more comfortable? Do you feel like you'll have enough time off?
 - Katie & I also put together an [Incident Playbook](https://dosomething.github.io/communal-docs/Incidents/) which outlines reasonable steps to take when responding to an alert. If you have time, take a look & let us know if we missed anything!

### DIFF
--- a/007-on-call.md
+++ b/007-on-call.md
@@ -1,0 +1,43 @@
+# On-Call
+
+We should use [PagerDuty](http://pagerduty.com) to centralize our alerting in one place & manage an on-call rotation.
+
+**Discussion:** https://github.com/DoSomething/rfcs/pull/7
+
+## Problem
+
+We rely on a plethora of monitoring services ([New Relic](https://newrelic.com), [Papertrail](https://papertrailapp.com), [Ghost Inspector](https://ghostinspector.com), [Runscope](https://www.runscope.com), [MongoDB Atlas alerts](https://docs.atlas.mongodb.com/monitoring-alerts/), and [Amazon CloudWatch](https://aws.amazon.com/cloudwatch/)) that all have their own configuration & notification methods.
+
+It's also not clear who's on-call at any given time, and so _everyone's_ always effectively on-call. This isn't ideal, since it means that people always need to be ready to respond to alerting and don't ever have "real" time off.
+
+## Proposal
+
+We should use [PagerDuty](http://pagerduty.com) to centralize our alerting in one place & manage an on-call rotation.
+
+### Centralized Alerts
+
+PagerDuty allows us to centralize notifications for all of our monitoring services in one place. This reduces the chance that we forget to add someone to the notification rules for a particular service.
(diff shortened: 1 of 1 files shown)


DFurnes:
To help visualize the "first responder" on-call rotation, here's how January would look:

Screen Shot 2019-12-11 at 11 21 41 AM

So if an alert went off on January 3rd, Diego & Aaron would be on-call to respond to it. If an alert went off on January 15th, the alert would instead be routed to Vanessa & me.

mshmsh5000:
This is a really readable and thorough first pass!

A note and two questions:

1. The Incident Playbook link above 404s. This works: https://dosomething.github.io/communal-docs/Incidents/
2. How might we test this in the context of a fire drill? What is the LOE of breaking a service in QA and feeding the incident into PD?
3. Do all on-call engineers for a specific service have the access they need? I'm thinking particularly of Apollo, which just started charging much more for each user, but there may be others where we don't by default have accounts for every engineer.

DFurnes:
Whoops, fixed that link!

> How might we test this in the context of a fire drill? What is the LOE of breaking a service in QA and feeding the incident into PD?

Yeah, I'd definitely like to have alert policies set up in QA so we can "run through" the fire drill as if it were a real incident. It might be a good time to revisit whether we can manage all this with Terraform... at least for [things](https://www.terraform.io/docs/providers/pagerduty/index.html) [that](https://www.terraform.io/docs/providers/runscope/index.html) [have](https://www.terraform.io/docs/providers/newrelic/index.html) providers.

> Do all on-call engineers for a specific service have the access they need? I'm thinking particularly of Apollo, which just started charging much more for each user, but there may be others where we don't by default have accounts for every engineer.

Yeah, it's such a bummer that they changed their pricing. I can't think of the last time we've actually triggered an alert from Apollo (knock on wood) so maybe that's something we can hold off on for now, and just auto-escalate those to myself or Katie if it does occur.

I don't think we have everyone set up to view Mongo & AWS alerts either (although, again, awfully unlikely to actually get an alert). I can set up accounts for anyone who's missing those.

mirie (approved):

rapala61 (approved):

mendelB (approved):

mshmsh5000 (approved):
👍

DFurnes:
That's four approvals, which means this is ready to merge! If anyone has feedback as we continue our inaugural on-call rotation in 2020, please don't hesitate to leave it here or in Mai's anonymous feedback box! 😸

PR #10, opened 2020-04-08
Add 'Server-Side Rendering' RFC.

DFurnes:
This pull request introduces the ["Server-Side Rendering" RFC](https://github.com/DoSomething/rfcs/blob/ssr/010-server-side-rendering.md):

> We should use [Next.js](https://nextjs.org) and [Heroku](https://www.heroku.com) to power server-side rendering for Phoenix.

I tried to summarize the problems and solution as concisely as I could in the first two sections. 

The "Alternatives & Comparison" section is significantly more verbose, but hopefully helps fill in the reasoning for how I got there. If it's too much to read in one sitting (fair!), it's broken up into topics that are linked from the relevant parts of the proposal itself.

# Open Questions
- In general, do you think these are the right trade-offs? What concerns do you have?
- Any problems you think we might run into that I've missed?

### DIFF
--- a/010-server-side-rendering.md
+++ b/010-server-side-rendering.md
@@ -66,7 +66,7 @@ At first glance, we're just moving the same work from the client to the (admitte
 
 #### Static Rendering & On-Demand Rendering
 
-Next.js supports [opt-in static generation](https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation) alongside on-demand rendered routes. This means that pages that are completely static (say, an article page) can be rendered to HTML at build-time while other pages continue to be rendered on-demand. Unfortunately, this [doesn't work with Apollo's server-side rendering](#) & so we're stuck with a choice between static generation & pre-rendering colocated queries.
+Next.js supports [opt-in static generation](https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation) alongside on-demand rendered routes. This means that pages that are completely static (say, an article page) can be rendered to HTML at build-time while other pages continue to be rendered on-demand. Unfortunately, this [doesn't work with Apollo's server-side rendering](https://github.com/DoSomething/phoenix-next/pull/1970#issuecomment-623726651) & so we're stuck with a choice between static generation & pre-rendering colocated queries.
 
 #### Further Rendering Optimizations
 
--- a/010-server-side-rendering.md
+++ b/010-server-side-rendering.md
@@ -11,7 +11,7 @@ Our front-end application, Phoenix, is not as scalable, simple, or performant as
 - Because Phoenix runs on traditional server processes, our **ability to scale is limited** (for authenticated users). If we’re unable to gracefully scale with traffic spikes, we might lose engagement at a key time.
 - The application is split between a traditional backend server, written in [Laravel](https://laravel.com), and a client-side application, written in [React](https://reactjs.org). This **increases context switching for developers**, slowing down development.
 - Because the application is almost completely rendered on the client, we see **reduced performance on slower devices** (like old phones). This may reduce engagement, since nobody likes waiting.
-- Finally, we've been limited in what we can share on social networks (whose scrapers do not execute client-side code) since any social metadata must be fetched and rendered by a separate server-side process.
+- Finally, we've been limited in what we can share on social networks (whose scrapers do not execute client-side code) since any social metadata must be fetched and rendered by a separate server-side process. This also impacts built-in web functionality, like using `#anchor-links` to link directly to a section of a page.
 
 Despite these problems, we're generally happy with what we've built in Phoenix & don't want to consider a full re-write. We know that approach [often turns into a quagmire](https://www.joelonsoftware.com/2000/04/06/things-you-should-never-do-part-i/) and so whatever we choose, we want to be able to adopt it incrementally with our existing codebase (sorry [Nuxt.js](https://vuejs.org), [Svelte](https://svelte.dev) and [Livewire](https://laravel-livewire.com)!)
 
--- a/010-server-side-rendering.md
+++ b/010-server-side-rendering.md
@@ -1,0 +1,205 @@
+# Server-Side Rendering
+
+We should use [Next.js](https://nextjs.org) and [Heroku](https://www.heroku.com) to power server-side rendering for Phoenix.
+
+**Discussion:** https://github.com/DoSomething/rfcs/pull/10
+
+## Problem
+
+Our front-end application, Phoenix, is not as scalable, simple, or performant as we want:
+
+- Because Phoenix runs on traditional server processes, our **ability to scale is limited** (for authenticated users). If we’re unable to gracefully scale with traffic spikes, we might lose engagement at a key time.
+- The application is split between a traditional backend server, written in [Laravel](https://laravel.com), and a client-side application, written in [React](https://reactjs.org). This **increases context switching for developers**, slowing down development.
+- Because the application is almost completely rendered on the client, we see **reduced performance on slower devices** (like old phones). This may reduce engagement, since nobody likes waiting.
+- Finally, we've been limited in what we can share on social networks (whose scrapers do not execute client-side code) since any social metadata must be fetched and rendered by a separate server-side process.
+
+Despite these problems, we're generally happy with what we've built in Phoenix & don't want to consider a full re-write. We know that approach [often turns into a quagmire](https://www.joelonsoftware.com/2000/04/06/things-you-should-never-do-part-i/) and so whatever we choose, we want to be able to adopt it incrementally with our existing codebase (sorry [Nuxt.js](https://vuejs.org), [Svelte](https://svelte.dev) and [Livewire](https://laravel-livewire.com)!)
+
+## Proposal
+
+We should use [Next.js](https://nextjs.org) and [Heroku](https://www.heroku.com) to power server-side rendering for Phoenix. They allow us to statically generate markup where possible (content and marketing pages), and [**render pages on-demand where it's not**](#static-vs-on-demand-rendering). This gives us the "best of both worlds", and allows us to incrementally optimize our site without a big architectural shift.
+
+We should [**keep our existing Apollo GraphQL-based data layer**](#data-layer), because it allows us to keep data requirements neatly co-located with usage & we don't want to be stuck re-writing tons of components. Similarly, I like that Next.js allows us to [**co-locate route data-fetching with components rather than in a monolithic "build" script**](#routing-layer).
+
+While all of these frameworks offer a [**great out-of-the-box developer experience**](#developer-experience), I found that Next.js aligned best with our existing technologies. It's also been [**battle-tested for large web applications**](#ecosystem--adoption) (like [Hulu](https://medium.com/hulu-tech-blog/introducing-hulus-modern-web-stack-d0be4cf4f439) and [Twitch](https://twitter.com/rauchg/status/1069683668132581376)), and not just simpler content/marketing sites.
+
+While [JAMStack](https://jamstack.wtf)-centric hosting providers (like [ZEIT Now](https://zeit.co) and [Netlify](https://www.netlify.com)) offer incredibly smooth setup, they're also very opinionated. For now, I'd recommend we [**continue to host on Heroku and Fastly**](#hosting) for a smoother transition.
+
+See [DoSomething/phoenix-next#1970](https://github.com/DoSomething/phoenix-next/pull/1970) for a working proof-of-concept.
+
+<br/>
+<p align="center">
+:dizzy:
+</p>
+<br/>
+
+## Alternatives & Comparison
+
+The two most popular React.js frameworks are definitely [Gatsby](https://www.gatsbyjs.org) and [Next.js](https://nextjs.org).
+
+I also looked into [RedwoodJS](https://redwoodjs.com), a new serverless web application framework (from Tom Preston-Warner, of [Jekyll](https://jekyllrb.com) fame). It's far too early to recommend adopting in production, but it has a lovely [philosophy](https://github.com/redwoodjs/redwood#the-redwood-philosophy), some very smart conventions ([cells](https://redwoodjs.com/tutorial/cells), omg!) and a [fantastic tutorial](https://redwoodjs.com/tutorial/welcome-to-redwood.html). Definitely worth keeping an eye on!
+
+We could [follow the Warren campaign's approach](https://dev.to/itsjoekent/server-side-rendering-react-in-realtime-without-melting-your-servers-21ej) and build our own Lambda/Node.js renderer using React's [`renderToString` API](https://reactjs.org/docs/react-dom-server.html) or [Puppeteer](https://developers.google.com/web/tools/puppeteer/articles/ssr). That would allow us to build a tool that perfectly matches our needs, but we'd also be "blazing our own trail" and losing many of the benefits of the vibrant communities that have adopted these frameworks! Because we're a small team, I've ruled that out as well.
+
+Regardless of which approach we chose, we'd keep most of the same application code – so we're really just comparing the "renderer", data-fetching layer, developer tooling, and community/documentation:
+
+### Static vs. On-Demand Rendering
+
+There are two different approaches we could take to server-rendering: either pre-rendering static HTML files (like the old days), or server-rendering pages on-demand & caching them with a CDN. In both cases, we'd be transforming React components into HTML – just at different points in the process.
+
+#### Static Rendering
+
+Turning Phoenix into a completely static HTML site is a very tempting idea since it drastically reduces the operational complexity of serving the site (it's just HTML!). Unfortunately, it accomplishes this by shifting the complexity into the build process. In addition to building all of our application code & styles, we'd _also_ need to build all of our content too.
+
+Gatsby is only able to statically render HTML, and so _everything_ we want in the server's response needs to be known at build time – requiring us to somewhat tediously tally up all of our content every time we build the site. I've heard this can get very slow (in some cases, hour-long builds!), although [Gatsby Build seems to help](https://www.gatsbyjs.org/blog/2020-01-27-announcing-gatsby-builds-and-reports/#builds-making-deploys-as-fast-as-our-websites)
+
+Ideally, we could build each page on-demand as it changes so that content changes would remain fast. We'd still need to run a full build for app changes, though, and incremental builds for Gatsby [are not yet shipped](https://www.gatsbyjs.org/blog/2020-01-27-announcing-gatsby-builds-and-reports/#are-builds-the-same-thing-as-incremental-builds).
+
+
+#### On-Demand Rendering
+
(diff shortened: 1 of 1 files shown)


DFurnes:
One of the things I've omitted for now is how we'd handle authentication without a traditional backend. I think we'll likely want to [use the Authorization Grant & PKCE flow](https://github.com/DoSomething/northstar/issues/840) (since a client-side application can't very well keep a "secret"), but that'll need some more investigation.

This shouldn't be a blocker for getting some high-impact pages (like the article and referral) pages to be server-rendered next quarter (since they don't need authentication), but it's something we'll have to consider before we can get everything moved over.

010-server-side-rendering.md
```diff
@@ -0,0 +1,205 @@
+# Server-Side Rendering
+
+We should use [Next.js](https://nextjs.org) and [Heroku](https://www.heroku.com) to power server-side rendering for Phoenix.
+
+**Discussion:** https://github.com/DoSomething/rfcs/pull/10
+
+## Problem
+
+Our front-end application, Phoenix, is not as scalable, simple, or performant as we want:
```
mendelB, line 9: Is anchor linking enough of a problem to make this list? I feel like that was a really nice win when we introduced linking to specific blocks on campaign pages etc. which we lost with dynamically loaded blocks.

010-server-side-rendering.md
```diff
@@ -0,0 +1,205 @@
...
+
+- Because Phoenix runs on traditional server processes, our **ability to scale is limited** (for authenticated users). If we’re unable to gracefully scale with traffic spikes, we might lose engagement at a key time.
+- The application is split between a traditional backend server, written in [Laravel](https://laravel.com), and a client-side application, written in [React](https://reactjs.org). This **increases context switching for developers**, slowing down development.
+- Because the application is almost completely rendered on the client, we see **reduced performance on slower devices** (like old phones). This may reduce engagement, since nobody likes waiting.
+- Finally, we've been limited in what we can share on social networks (whose scrapers do not execute client-side code) since any social metadata must be fetched and rendered by a separate server-side process.
```
mendelB: Playing devil's advocate, would it be fair to argue that all of these issues are addressable without resorting to server-side rendering? (e.g. completely moving our backend 3rd party API calls to front-end/GraphQL and coming up with some general strategy for SEO and scraping.) 

Is there a general argument specifically for server-side rendering notwithstanding us solving these issues on their own? (Or, is this counter-argument invalid - and these issues _are_ still big enough to warrant the switch e.g. I didn't address scaling.)

010-server-side-rendering.md
```diff
@@ -0,0 +1,205 @@
...
+
+Despite these problems, we're generally happy with what we've built in Phoenix & don't want to consider a full re-write. We know that approach [often turns into a quagmire](https://www.joelonsoftware.com/2000/04/06/things-you-should-never-do-part-i/) and so whatever we choose, we want to be able to adopt it incrementally with our existing codebase (sorry [Nuxt.js](https://vuejs.org), [Svelte](https://svelte.dev) and [Livewire](https://laravel-livewire.com)!)
```
mendelB: *The `Nuxt.js` link links directly to the Vue library itself and not to https://nuxtjs.org/

mendelB (commented):
I haven't made it too far into the Alternatives & Comparison yet but left a few (hopefully helpful) comments for now!

010-server-side-rendering.md
```diff
@@ -0,0 +1,205 @@
...
+
+At first glance, we're just moving the same work from the client to the (admittedly, often faster) server. However, we've got a trick up our sleeve – caching! Once we've rendered a page once, we can store it in a public cache (on [Fastly](https://www.fastly.com) or [Zeit's CDN](https://zeit.co/smart-cdn)), making subsequent visits to that page "render" instantaneously.
+
+#### Static Rendering & On-Demand Rendering
+
+Next.js supports [opt-in static generation](https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation) alongside on-demand rendered routes. This means that pages that are completely static (say, an article page) can be rendered to HTML at build-time while other pages continue to be rendered on-demand. Unfortunately, this [doesn't work with Apollo's server-side rendering](#) & so we're stuck with a choice between static generation & pre-rendering colocated queries.
```
mendelB: 👾 It looks like the `doesn't work with Apollo's server-side rendering` link is missing.

mendelB: I didn't really understand the Apollo server-side rendering issue -- does it not work with on-demand rendering? So we'd have to go with static rendering?

010-server-side-rendering.md
DFurnes: Oh, good call! I think that definitely belongs here.

010-server-side-rendering.md
DFurnes: Whoops!

010-server-side-rendering.md
DFurnes: Yeah, the first two could absolutely be solved by moving more logic to the client-side (and having Laravel simply [render that empty `app` view](https://github.com/DoSomething/phoenix-next/blob/c04752094f8234c31208c615b4f7c61dd8d8ecc6/routes/web.php#L60)). The more we do that, though, the more the second two problems suffer (rendering performance & un-scrapable social metadata).

I think what I find most appealing about tackling server-rendering is that it lets us make meaningful improvements for all of these pain points in one fell swoop.

DFurnes: I think the final problem perhaps makes the cleanest case for server-side rendering.

Since Facebook & Twitter don't run client-side code, we have to provide those metatags in the server's HTML response. This means we either:
1. can't include dynamic metadata on client-rendered pages (like company/hub pages)
2. have to make all of our requests on the server, which introduces that awkward conceptual split & prevents us from using client-side routing for better performance.
3. have to double-up those requests (with a server-side Contentful API request & a client-side GraphQL request), which adds a meaningful performance penalty.

The way we'd so far tried to improve developer experience & scalability (moving to client-side rendering) has a direct negative effect on performance & "scrapability".

010-server-side-rendering.md
DFurnes: Gah yeah! I meant to write up more details on that in my proof-of-concept PR and link here, but totally forgot! I'll make myself a note to do that later today.

The TLDR is that when we use Apollo's ability to "render the tree" and fetch up any queries within a view, we aren't able to "optimize" to a statically-rendered page; so we have to choose an "either or" for those two approaches (for now, at least).

010-server-side-rendering.md
mendelB: Ah yeah, I'd agree that the final problem indeed made the strongest impression on me. Particularly since we've (as you well know lol) [had to deal with this](https://github.com/DoSomething/phoenix-next/pull/1979) pain point directly.

And I quite like the 'one fell swoop' point! Altogether, color me convinced. and thanks for clarifying!

010-server-side-rendering.md
DFurnes: Added this in df9f5f0.

010-server-side-rendering.md
DFurnes: Fixed in ea581c0!

010-server-side-rendering.md
```diff
@@ -0,0 +1,205 @@
...
+
+I also looked into [RedwoodJS](https://redwoodjs.com), a new serverless web application framework (from Tom Preston-Warner, of [Jekyll](https://jekyllrb.com) fame). It's far too early to recommend adopting in production, but it has a lovely [philosophy](https://github.com/redwoodjs/redwood#the-redwood-philosophy), some very smart conventions ([cells](https://redwoodjs.com/tutorial/cells), omg!) and a [fantastic tutorial](https://redwoodjs.com/tutorial/welcome-to-redwood.html). Definitely worth keeping an eye on!
+
+We could [follow the Warren campaign's approach](https://dev.to/itsjoekent/server-side-rendering-react-in-realtime-without-melting-your-servers-21ej) and build our own Lambda/Node.js renderer using React's [`renderToString` API](https://reactjs.org/docs/react-dom-server.html) or [Puppeteer](https://developers.google.com/web/tools/puppeteer/articles/ssr). That would allow us to build a tool that perfectly matches our needs, but we'd also be "blazing our own trail" and losing many of the benefits of the vibrant communities that have adopted these frameworks! Because we're a small team, I've ruled that out as well.
```
weerd, line 42: > That would allow us to build a tool that perfectly matches our needs, but we'd also be "blazing our own trail" and losing many of the benefits of the vibrant communities that have adopted these frameworks! Because we're a small team, I've ruled that out as well.

This definitely feels like the right call! 👍

010-server-side-rendering.md
```diff
@@ -0,0 +1,205 @@
...
+
+#### Static Rendering & On-Demand Rendering
+
+Next.js supports [opt-in static generation](https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation) alongside on-demand rendered routes. This means that pages that are completely static (say, an article page) can be rendered to HTML at build-time while other pages continue to be rendered on-demand. Unfortunately, this [doesn't work with Apollo's server-side rendering](#) & so we're stuck with a choice between static generation & pre-rendering colocated queries.
+
+#### Further Rendering Optimizations
+
+However, what about things like displaying a user's personalized submission gallery? We don't want a tiny bit of personalization to prevent an entire page from being cached. In those cases, we can [skip particular GraphQL queries](https://www.apollographql.com/docs/react/performance/server-side-rendering/#skipping-queries-for-ssr) when server-rendering. In this example, that allows us to render the rest of the campaign template & any user-specific elements would simply "pop" in on the client-side!
```
weerd, line 73: Noice 👌

010-server-side-rendering.md
```diff
@@ -0,0 +1,205 @@
...
+
+The biggest downside is that stitching an existing GraphQL API (like [graphql.dosomething.org](https://graphql.dosomething.org)) into Gatsby's schema requires prefixing query and type names (e.g. `Post` would have to become `DSPost`). Gatsby is also only able to server-render queries in top-level pages, or in components that use its own custom [`useStatic`](https://www.gatsbyjs.org/docs/static-query/) hook. Both of these limitations would require significant re-writes of existing code.
+
+In Next.js, we can fetch data using Apollo's [server-side rendering support](https://www.apollographql.com/docs/react/performance/server-side-rendering/). This allows us to server-render an entire page of GraphQL-backed components with no changes to our existing data fetching logic! And anything that doesn't make sense to query via GraphQL can be fetched with [`getServerSideProps`](https://nextjs.org/docs/basic-features/data-fetching).
```
weerd, line 83: 🎉

weerd (approved):
✅ 👍 This looks great to me!

010-server-side-rendering.md
katiecrane: I think I'm still confused about this Apollo piece - it can't do static generation but it does generally work with Next.js?

010-server-side-rendering.md
katiecrane: So this means we'd have to tag any queries involving user id or anything user specific to have that `ssr: false` piece in order for pages using those queries to be cache-able? I imagine we'd mostlyyyy want to be caching data we get from Contentful is that correct? I'm glad there is a way to handle this!

010-server-side-rendering.md
mendelB: > This means that pages that are completely static (say, an article page) can be rendered to HTML at build-time

How does this type of rendering work with content changes, is it the on-demand rendering you referenced above?

mendelB (approved):
I'm super excited about this!

Next.js definitely sounds like the best bet considering our specific circumstances and I'm really excited about the developer features we'd gain! Heroku certainly seems like a no-brainer for now.

I'm particularly intrigued by the caching considerations this would introduce and am overall super psyched to dig into the new (old!) server rendering world...

Thanks for (yet another) awesome write up!

lindsayrmaher (approved):
I'm excited for this! I can't say I needed a ton of convincing, I liked Next.js when using it to prototype other applications in the past and it definitely seems to suit our needs pretty efficiently. Also _always_ for JS full stack babyyyyyy 😎

My biggest questions are also around the Apollo-Querying that a couple other people have left some notes on, but it sounds as though it's not a big enough blocker that we would have to completely change our dev flows. 

🤸‍♀️

010-server-side-rendering.md
```diff
@@ -0,0 +1,205 @@
...
+
+### Routing Layer
+
+Both Gatsby and Next.js provide their own built-in routing systems. Both frameworks use the filesystem to add simple routes, but they differ in how they handle dynamic routes (like our `us/facts/{fact}` pages):
+
+Next.js uses declarative [dynamic routes](https://nextjs.org/docs/routing/dynamic-routes) – you say what your routes look like (e.g. `us/facts/[fact].js`) as the page component itself. It can then be rendered on-demand as a request comes in, or (optionally) pre-rendered by using the [`getStaticPaths`](https://nextjs.org/docs/basic-features/data-fetching#getstaticpaths-static-generation) hint. This allows us to keep all of a page's concerns in one place:
```
VCGibbs, line 89: awesome!

VCGibbs (approved):
So far I really like Next.js! Also this RFC is very thorough and informative. I really like the trajectory in efforts to incorporate optimized technology on our platform.  I like to minimalistic aspect of next.js, the server-side rendering, lazy-loading, etc. I think it will definitely improve loading time. I'm assuming this will also enhance Phoenix's security? what about testing? Will anything change in regards to debugging, using chrome dev tools or react dev tools?

010-server-side-rendering.md
```diff
@@ -0,0 +1,205 @@
...
+
+### Developer Experience
+
+Developer experience for both frameworks is great, and offer big improvements over what we're used to!
+
+More than anything else, I'm looking forward to being able to follow the framework's established conventions (where previously we've had to forge our own path). Because these frameworks have a standardized approach to routing, data fetching, and tooling, we'll spend less time [bikeshedding](https://en.wikipedia.org/wiki/Law_of_triviality) and have a much easier time onboarding new engineers (since if you know the framework, you'll know how we use it).
+
+These conventions also enable some exciting features. For example, both frameworks support "hot reloading" out of the box – which means that changes are _immediately_ reflected in the browser without reloading the page! It feels like magic. They also both display compilation errors in the browser (so no more tabbing over to the terminal to check if the build failed):
```
katiecrane, line 161: HOT RELOADING AHHHHH!!! That's so cool!

katiecrane: AND compilation errors in the browser is really excellent too 🎉

010-server-side-rendering.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Drawbacks
+
+While adding server-side rendering reduces context switching between two codebases (PHP and React), it introduces its own overhead since we can now render one codebase in two environments (Node.js for SSR & browser). Developers will need to be aware of these two differing "modes" a component may be rendered in, and re-write code appropriately (such as being careful about `window` usage).
+
+Server-side rendering is also still rather resource intensive, and we'll be heavily relying on caching to avoid overloading our backend. Luckily, we can always fall back to client-side rendering if necessary, and can use [`stale-while-revalidate`](https://www.fastly.com/blog/stale-while-revalidate-stale-if-error-available-today) to make slower requests to the backend without slowing down user-facing requests.
```
katiecrane, line 201: `stale-while-revalidate` seems like a really neat idea!

> Luckily, we can always fall back to client-side rendering if necessary

Can you explain this bit a little more? I thought this was more of a switchover, how could we fall back?

katiecrane (approved):
I like this! I left some questions that are a bit in the weeds, but overall I think you covered everything!

aaronschachter (approved):
Great writeup! This will be really useful to reference as other new tools appear / existing tools evolve.

010-server-side-rendering.md
DFurnes: We'd have to [re-trigger a build](https://www.contentful.com/developers/docs/tutorials/general/automate-site-builds-with-webhooks/) whenever content changes in order to rebuild all these static HTML pages. (We could perhaps consider using the [`fallback`](https://nextjs.org/docs/basic-features/data-fetching#fallback-false) option for `getStaticPaths` to get the "best of both worlds" and lazily re-render those static pages once they get a request if these builds get slow...)

There's also [Preview Mode](https://nextjs.org/docs/advanced-features/preview-mode), which would let editors bypass static generation (for viewing things on [preview.dosomething.org](https://preview.dosomething.org) before they're actually published).

010-server-side-rendering.md
DFurnes: Yep! I think that's exactly how we'd handle this – we'd server-render and cache anything that's generic enough that most or all users could benefit from it, and client-render anything that's specific to a particular person or would otherwise cache poorly (like reaction counters in the gallery perhaps).

010-server-side-rendering.md
DFurnes: There should be nothing stopping any of these Next.js pages from being rendered client-side (and in fact, that's what's happening on subsequent navigation within the site). Server-side rendering is more of an enhancement where we save the client some work by shipping them a "finished product" the first time they visit a page!

So in the unlikely chance that server-side rendering was really taxing our servers, we could always "disable" it for a subset of pages & just ship the same sort of [application shell](https://github.com/DoSomething/phoenix-next/blob/master/resources/views/app.blade.php) (and render it all client-side). It'd be a more extreme version of the same way we'd opt-out of server-rendering some Apollo components with `ssr: false`.

DFurnes:
> I'm assuming this will also enhance Phoenix's security? what about testing? Will anything change in regards to debugging, using chrome dev tools or react dev tools?

By moving more in the "JAMStack" direction (by replacing servers that run our own code with managed services like Contentful), we'll definitely reduce our attack surface. And we'll eventually have one less Laravel site to worry about keeping updates & patched, which is great.

This should hopefully make testing a lot easier, since one of our biggest pain points when writing Cypress tests is that portions of our application logic are still locked away in PHP code (that we can't easily mock during test runs).

And Chrome & React dev tools will continue to work exactly the same way!

DFurnes:
> This will be really useful to reference as other new tools appear / existing tools evolve.

Agreed!! I'm definitely keeping an eye on [Redwood](https://redwoodjs.com), and just saw [Remix](https://remix.run) the other day which also looks super promising! The nice thing is that a lot of these tools take similar approaches (and are, ultimately, all React apps), so it should be easier to migrate between frameworks later on.

DFurnes:
Thank you all for the fantastic comments and questions – both here and in our roundtable demo and Q&A last week. Really excited to kick off the first phase of this work later this quarter! 👯

010-server-side-rendering.md
DFurnes: Finally, added the [promised write-up](https://github.com/DoSomething/phoenix-next/pull/1970#issuecomment-623726651) on the issues I ran into with `getStaticProps` (for static generation) and Apollo's server-side rendering support in 323f31f.

PR #11, opened 2020-08-11
Add RFC for PHP Code Formatting.

weerd:
This pull request introduces the **"PHP Code Formatting" RFC**:

> We should implement the use of both the [Prettier PHP Plugin](https://github.com/prettier/plugin-php) for prettifying the look of our PHP code and the [PHP Coding Standards Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) tool for ensuring the resulting PHP code is written well, following specified code quality rules.


011-php-code-formatting.md
```diff
@@ -0,0 +1,94 @@
...
+
+While there is going to be some natural overlap between the two tools and how they format code style, we can definitely benefit from where the two tools differ in what they are meant to fix, to get the best of both worlds.
+
+As the [Prettier docs explain](https://prettier.io/docs/en/comparison.html), Prettier's strength lies in removing the stress of how the code "looks", and ensures it is formatted the same across a team of developers. However, it does nothing to help with code quality rules and this is where PHP CS Fixer can fill that gap, and replace the role of what StyleCI did in a much more convenient and useful way.
+
+### Prettier PHP Plugin
+
+[Prettier PHP Plugin](https://github.com/prettier/plugin-php) is an opinionated code formatter that will rewrite outputted PHP code that conforms to a consistent style, aiming for a line length print width of 80 characters.
+
+It helps developers eliminate any concerns regarding how their PHP code looks because Prettier takes care of that for them.
```
DFurnes: One of the things I love about Prettier is how little configuration it allows - so there's less opportunity to bike-shed on which style rule we prefer! Do you think that's worth calling out as an advantage here?

011-php-code-formatting.md
```diff
@@ -0,0 +1,94 @@
...
+
+For a quick visual comparison, here is the same file edited separately by the two tools:
+
+#### Prettier Changes
+
+You can see in these [Prettier file changes](https://github.com/DoSomething/rogue/pull/1094/files#diff-5c9547d315e53dd95227ca9984b56d06), it edited the line-lengths of the code.
+
+#### PHP CS Fixer Changes
+
+You can see in these [PHP CS Fixer file changes](https://github.com/DoSomething/rogue/pull/1092/files#diff-5c9547d315e53dd95227ca9984b56d06), it sorted the class imports to be alphabetical and removed the spacing around string concatenation.
```
DFurnes: I think the string concatenation rule is a good example of something we'll want to let Prettier handle, since it seems more like a "formatting" rule to me. I also don't believe we can customize [Prettier's behavior](https://loilo.github.io/prettier-php-playground/#N4IgDgTgpgLjCWUIHkwIPYDsDOIBcok8mMA6vACYwAW+AHAAwA0IMAhgEblW14AsLAK7YoAFU648AMzYAbES2zEA5rKgBFQehhR8M+VBZhqYAGpIlWfCACsAOj4gWMCG3iyVAYXQBbH2wAFAAkA-BdBQxAOVwBjKABlGABPNWswbAgAWgAmJxBoAEdBeGgA12V-PTkFEGIRCBgytgq2KoMAXxYoCngYdBQ0eCxJUDZsGDaRTtrMMEEJvBAAHgB+YzAAHUwtqBjqdAACAHIdcZUjuyPxiHO7QTAwJAAKI5isGLYYI4BKAG4QdpAA) on this one & wouldn't want the two tools to conflict.

011-php-code-formatting.md
```diff
@@ -0,0 +1,94 @@
...
+
+When StyleCI runs to check the PHP code style on a push to a branch, if it finds errors, the run fails with a link to details in the pull request. At times these details are very confusing to understand exactly what the recommended change is. Developers sometimes need to try a few edits, commit them and see if the styles pass on subsequent runs, but there is no way to know without pushing up code.
+
+StyleCI does offer to create pull requests on demand for these changes, but that workflow feels a bit disjointed given that a developer would like this all fixed in their initial pull request.
+
+It would be ideal to be able to know exactly what needs to be changed and confident that doing so has fixed the style issue; and not needing to push changes numerous times to get feedback from StyleCI on whether the changes pass or not.
```
DFurnes, line 19: This one is gonna be a game-changer. Those little :x:s always catch me at the last moment!!

DFurnes (approved):
Really excited about this!

011-php-code-formatting.md
weerd: Ah that's a good point. When I initially reviewed the results of having run Prettier on the codebase it seemed like the concatenation rule was something it didn't address (assuming it would remove spaces 😅)... I hadn't realized it didn't touch them BECAUSE it considered that correct! haha. But good to know my theory on "contradicting" rules was right!

I think my planned approach, if approved, will be to first merge in the Prettier PR and establish what Prettier normally fixes. Knowing that I can customize a ruleset for PHP CS Fixer to add to `weerd/php-style` that doesn't conflict called `prettier-laravel` or something.

011-php-code-formatting.md
```diff
@@ -0,0 +1,94 @@
...
+
+**Discussion:** <https://github.com/DoSomething/rfcs/pull/11>
+
+## Problem
+
+For most of our repositories that include PHP we currently depend on [StyleCI](https://styleci.io) for formatting our code.
+
+While it has served our needs for the most part, this tool is not as convenient as some other tools available, and at times has even been a bit confusing to use.
+
+The setup for StyleCI includes adding a GitHub repository to the list of projects it should "check" automatically for each "push" to a pull request branch. The settings can be customized on the StyleCI web interface but each project can also contain their own configuration in a `.styleci.yml` file. Having multiple locations for configuration settings can be a bit confusing; not always being apparent if a project is checked by StyleCI if the local config file is not present.
```
mendelB, line 13: I don't know if it's necessarily worth explicitly pointing out our [various](https://dosomething.slack.com/archives/C3ASB4204/p1572615964090600) [kerfuffles](https://dosomething.slack.com/archives/CUQMU4Q6B/p1596116997002700?thread_ts=1596116997.002700) with unexpected default config changes which intersected with the configuration inconsistency we're pointing out here?

011-php-code-formatting.md
```diff
@@ -0,0 +1,94 @@
...
+
+You can see a [pull request](https://github.com/DoSomething/rogue/pull/1094) on our Rogue application with how the Prettier PHP Plugin is configured and how it proceeded to format the code.
+
+This [commit](https://github.com/DoSomething/rogue/pull/1094/commits/5e2a94b46168539a6346b6057efc2d915e444626) specifically shows how the Prettier PHP plugin is set up to run along side the configuration for our JavaScript code.
+
+This [commit](https://github.com/DoSomething/rogue/pull/1094/commits/88a8ed7d2e5f8d3e13e93c4840b2d389cd331614) specifically shows all the files changed when the `npm run format` command was executed.
+
+The way the plugin is configured, in daily use developers will not need to manually run the command to format their code since it will run as a `pre-commit` hook on any changed files.
+
+Additionally, some code editors (like Visual Studio Code) easily support running Prettier on file saves (refer to the code editor docs for documentation on setup).
```
mendelB, line 43: Sublime Text too Diego! I will not stand for this pretension and snobbery 😛

mendelB (approved):
👍  I've really enjoyed the benefits of ESList + Prettier in JS land (where I think we've managed to avoid the 'contradictions' drawback as far as I can tell) -- super excited about this!!! 

Thanks for including the examples, those were helpful for visualizing this!

011-php-code-formatting.md
weerd: I still ❤️  Sublime too! I use both. VSCode for large projects and ST for small and/or fast quick-and-dirty coding needs 🐎

011-php-code-formatting.md
weerd: @DFurnes  Added!

011-php-code-formatting.md
weerd: Ah good point @mendelB ! I added more info mentioning lack of version control and linked to those Slack instances! <https://github.com/DoSomething/rfcs/pull/11/files#diff-f0c9652f43e84660cd5f067a9f73c7ddR53>

lindsayrmaher (approved):
I approve!!! Esp excited about not relying on the styleci fails when code is pushed up 😅

011-php-code-formatting.md
DFurnes: That makes sense to me! 👍

011-php-code-formatting.md
katiecrane: I love that with it running on commits and automatically on saves this isn't something I have to think about!!

011-php-code-formatting.md
```diff
@@ -0,0 +1,94 @@
...
+
+This [commit](https://github.com/DoSomething/rogue/pull/1092/commits/d48ec9a1a759941c49c5e687a80b639e9d808fed) specifically shows all the files changed when the `composer format` command was executed within the Homestead vagrant environment.
+
+Since this tool is run as a composer command, based on most of our development setups, the command should be run within the Homestead vagrant box. In practice, developers can potentially run the command manually, but it is also setup to run within our Continuous Integration environment and it will flag any failures within a pull request (similar to how StyleCI behaves).
+
+However, once flagged in a pull request, to automatically fix these errors, developers can just run the `composer format` command in Homestead and then commit and push up their code, confident that the errors were addressed and the style checks should pass.
```
katiecrane, line 59: Is the main difference between this and StyleCI in practice then that when I push up a PR that gets an ❌ , I can run a command to automatically fix the issues?

011-php-code-formatting.md
```diff
@@ -0,0 +1,94 @@
...
+
+This [commit](https://github.com/DoSomething/rogue/commit/eeca6efb6b29df01e13e816b7d051e872b8cd2f7) specifically shows how the PHP CS Fixer tool is set up. The process is simplified by utilizing the [weerd/php-style](https://github.com/weerd/php-style) package which contains easily accessible rulesets for both generic PHP packages and Laravel projects. This package is also version controlled, so developers can rest easy knowing they can specify a version of the ruleset to use and they can opt-in to ruleset changes by adjusting their composer package version. This scenario has been a pain point for the team in the past that affected us on [more than one](https://dosomething.slack.com/archives/C3ASB4204/p1572615964090600) [occasion](https://dosomething.slack.com/archives/CUQMU4Q6B/p1596116997002700?thread_ts=1596116997.002700), where StyleCI made a change and affected all our repositories since their rules are not version controlled.
+
+This [commit](https://github.com/DoSomething/rogue/pull/1092/commits/d48ec9a1a759941c49c5e687a80b639e9d808fed) specifically shows all the files changed when the `composer format` command was executed within the Homestead vagrant environment.
```
katiecrane, line 55: Looks like mostly changing the order of import statements (from shortest to longest to...alphabetical?) and removing not needed spaces from string concatenation - are things things that Prettier can't handle?

011-php-code-formatting.md
```diff
@@ -0,0 +1,94 @@
...
+
+For a quick visual comparison, here is the same file edited separately by the two tools:
+
+#### Prettier Changes
+
+You can see in these [Prettier file changes](https://github.com/DoSomething/rogue/pull/1094/files#diff-5c9547d315e53dd95227ca9984b56d06), it edited the line-lengths of the code.
+
+#### PHP CS Fixer Changes
+
+You can see in these [PHP CS Fixer file changes](https://github.com/DoSomething/rogue/pull/1092/files#diff-5c9547d315e53dd95227ca9984b56d06), it sorted the class imports to be alphabetical and removed the spacing around string concatenation. This would be an example where we would want our PHP CS Fixer ruleset to not fix string concatenation, since Prettier styles them with a space around string concatenation.
```
katiecrane, line 89: I'm confused about this case - if I run `composer format` to fix the ❌ on my PR, won't that spacing change it made automatically be reverted when I save the files because Prettier will run on save?

011-php-code-formatting.md
weerd: Relates to [this discussion](https://github.com/DoSomething/rfcs/pull/11#discussion_r469318473) with Dave. I'm going to enable Prettier first, then PHP CS Fixer after to compare what rules to remove from PHP CS Fixer that Prettier should only handle. There's some overlap but aiming to have the PHP linting happening via PHP CS Fixer which Prettier doesn't handle (thus the comparison to having both Prettier and ESLint in our JS). Here's a lot of what PHP CS Fixer will handle among PSR-2 rules: https://github.com/weerd/php-style/blob/master/src/rules/laravel.php

011-php-code-formatting.md
weerd: Ya, you'll see the failure in the PR with an ❌, but instead of having to jump into StyleCI and interpret what it's telling you to change and hopefully understanding, with PHP CS Fixer, you just run `composer format` in your Homestead, push up the changes and PR should pass, easy-peasy 🍋 🗜

011-php-code-formatting.md
weerd: Ya, I mention this in the "Contradictions seciton:

> Another potential drawback is the possibility of contradicting fixes. Where depending on the order of operations, a style fix could keep getting edited back and forth between Prettier and PHP CS Fixer.

But I [discuss this](https://github.com/DoSomething/rfcs/pull/11#discussion_r469318473) and propose a way to address once implementing. Aiming to remove the rules in PHP CS Fixer that conflict with Prettier so that we cater to each of their strengths.

011-php-code-formatting.md
katiecrane: Ah so is the PHP CS Fixer fully customizable? I only remember reading that Prettier is not

011-php-code-formatting.md
katiecrane: GREAT emojis!!! Okay I dig that! Especially since it won't be for _everything_ because hopefully Prettier can handle most of it!

011-php-code-formatting.md
weerd: Oh I could have been clearer on that. I mention the use of rulesets and changes being version controlled, but should maybe clarify that it is customizable. Ideally just once and you're done with it.

katiecrane: Ah got it! I think that's why I was confused and thought we could get stuck in loops!

katiecrane (approved):
Good from me! I like having Prettier handle most things since it'll fix them automatically and then still being able to fix things if my PR fails just by running a command!

011-php-code-formatting.md
weerd: @katiecrane I [updated the language](https://github.com/DoSomething/rfcs/pull/11/commits/fb3a4602d953c8803af6eb71615b2c0650f9913a) to help clarify this! Thanks for pointing it out!

PR #14, opened 2020-10-30
Add 'Single Codebase' RFC.

DFurnes:
This pull request introduces the "Single Codebase" RFC:

> We should consolidate our services into a single repository, with a consolidated monolithic backend.

# Open Questions
 - We've done a lot of work over the years, and it's easy to fall into a trap of only thinking of situations that confirm my hunch! What situations would have been made more difficult if we were all working in a single repository/application?
 - Are there other situations where you think that combining Northstar, Rogue, and Chompy would have made work easier?
 - What services make sense to keep independent from the monolith?

### DIFF
--- a/014-single-codebase.md
+++ b/014-single-codebase.md
@@ -22,7 +22,7 @@ However, breaking our monolith into a service-oriented architecture introduced n
 
 ## Proposal
 
-We should consolidate our backend services (Northstar, Rogue, and Chompy) into **a single monolithic Laravel application.**
+We should consolidate our Laravel services (Aurora, Northstar, Rogue, and Chompy) into **a single monolithic application.**
 
 - This will **simplify architectural decisions** & enable us to iterate faster. Instead of sorting new functionality into strict bounded contexts up front, we'll be able to simply place new entities into a single backend context. Architectural decisions will be less binding, as it's easier to refactor within a single codebase than across network boundaries.
 
--- a/014-single-codebase.md
+++ b/014-single-codebase.md
@@ -1,0 +1,68 @@
+# Single Codebase
+
+We should consolidate our services into a single repository, with a consolidated monolithic backend.
+
+**Discussion:** https://github.com/DoSomething/rfcs/pull/14
+
+## Problem
+
+We rebuilt our old [Drupal 7 monolith](https://github.com/dosomething/website-2013) into a suite of services – [Northstar](https://github.com/DoSomething/northstar), [Rogue](https://github.com/DoSomething/rogue), [Phoenix](https://github.com/DoSomething/phoenix-next), etc. – between 2013-2019. We avoided [the quagmire of a total rewrite](https://www.joelonsoftware.com/2000/04/06/things-you-should-never-do-part-i/) by extracting functionality [piece-by-piece](https://martinfowler.com/bliki/StranglerFigApplication.html) & have been generally happy with what we've built since then.
+
+However, breaking our monolith into a service-oriented architecture introduced new issues of its own:
+
+- **Blurred boundaries:** The [bounded contexts](https://martinfowler.com/bliki/BoundedContext.html) for each of our services (especially our identity & activity APIs) can be blurry. For example, is joining a club part of a user's identity or an action that they take? We're often forced to make this decision early in a feature's development process, before the technical work has fully taken shape.
+
+- **Technical overhead:** Logic that straddles multiple domains is also inevitably more complicated, requiring network requests for what could otherwise be a function call or `JOIN` statement. It makes traditionally simple logic into distributed systems problems where we have to decide whether we'll trade consistency for performance or vice versa.
+
+- **Testing limitations:** We know we get the best bang for our buck from [high-level integration tests](https://github.com/DoSomething/rfcs/blob/master/004-testing.md), but separating into services forces us to use mocking extensively because the majority of our application (e.g. users in Rogue, or users & posts in Phoenix) is always an external dependency in any given test! This places a large burden on our intra-service contracts.
+
+- **Knowledge silos:** Finally, this separation reinforces silos (e.g. "you have to talk to Dave about that app" or "that's X application's responsibility, not ours"). We have a smaller team today than when we first started splitting out into a service oriented architecture, and it's important that we can all understand & comfortably work on the whole system.
+
+
+
+## Proposal
+
+We should consolidate our backend services (Northstar, Rogue, and Chompy) into **a single monolithic Laravel application.**
+
+- This will **simplify architectural decisions** & enable us to iterate faster. Instead of sorting new functionality into strict bounded contexts up front, we'll be able to simply place new entities into a single backend context. Architectural decisions will be less binding, as it's easier to refactor within a single codebase than across network boundaries.
+
+- It will also **simplify adding behavior that straddles contexts**, like clubs, by replacing network requests with function calls. We'll also be able to ensure better consistency by using more transactional validation.
+
+- Keeping more of our business entities in the same domain allows **easier integration testing**, since we can seed (and otherwise directly reference) data that would currently live across API boundaries, like user IDs & clubs.
+
+- We have less applications to keep up-to-date with security & framework upgrades!
+
+For services that may not make sense to fold into the monolith, like [GraphQL](https://github.com/DoSomething/graphql/) or [Gambit](https://github.com/DoSomething/gambit/), we should still collaborate on them [**as part of a single monorepo**](https://docs.google.com/document/d/1a3e0Wpv9qeRscspdbN4wVR4Chz5oJFb9HH69x2pngSg/edit), side-by-side with our consolidated backend application. This allows us to retain the technical benefits we've gotten from separating out these services, while collaborating in one place.
+
+
+
+## Drawbacks/Alternatives
+
+#### Okay, but on the flip side – why _did_ we move from a monolith to (micro)services?
+
+Our Drupal application had a lot of intertwined modules that could be hard to follow, due to Drupal 6's "hooks" based architecture. We wanted to use service boundaries to force a "separation of concerns" that we thought would keep our codebase tidier.
+
+At the time, microservices were the architecture du jour & we followed the success stories of teams many times our size, from Spotify to Google, that promised that smaller services would increase velocity and allow teams to ship faster. At first, we had roughly one team of developers per application ("Rocket" for Phoenix & Northstar, "Bleed" for Rogue, and "O'Doyle" for Gambit & Blink) and it really did increase velocity! (That velocity slowed as a we began to build more cross-cutting features later on.)
+
+We also found that splitting our monolithic application into services allowed us to choose "the best tool for the job", such as using Node.js for our chatbot. (We love PHP at Do Something, but we know it's not appropriate for every task).
+
+Finally, separating logic into individual services allowed us to scale them more granularly – for example, spinning up our SMS service during a broadcast, but not our front-end website. (This became less effective over time, as we now scale up Gambit, Northstar, and Rogue in tandem every time we run a broadcast.) 
+
+I don't think these were the wrong choices at the time, especially given the information we had! Given what we've learned over the past five years, though, I think we would not make the same decision if we started down this path today.
+
+
+#### Why would we not want to make this change now?
+
+First of all, it takes time! We'd be betting that the time we'd take to rearchitect these applications now would be outweighed by the increase in velocity we'd see as a result. Luckily, we're in a bit of a "lull" at the end of the year & this feels like an opportune time to "reset" and set ourselves up for success in the new year.
+
+We might also find that if we were to grow the team aggressively in the future, we'd once again start tripping over each other in a single codebase. It's hard to predict the future, but it doesn't seem likely at the moment that will be the case.
+
+
+
+## Additional Resources
+
+A lot of inspiration for this proposal comes from DHH's [The Majestic Monolith](https://m.signalvnoise.com/the-majestic-monolith/) (2015) and [The Majestic Monolith Can Become The Citadel](https://m.signalvnoise.com/the-majestic-monolith-can-become-the-citadel/) (2020), which detail Basecamp's "majestic" monolithic architecture. They have a small team of 12 developers, and most of their functionality lives in a single codebase (with a few extracted services where necessary).
+
+Another great case study is Segment's [Goodbye Microservices](https://segment.com/blog/goodbye-microservices/) (2018), which explains why they transitioned their famous microservice architecture back to a monolith. They cite improved productivity, testability, and reduced operational overhead.
+
+Finally, in [MonolithFirst](https://martinfowler.com/bliki/MonolithFirst.html) (2015), Martin Fowler argues that projects should start as a monolith and only be broken up once the system's bounded contexts are solidly established. At that point, microservices can be peeled off around the edges.


014-single-codebase.md
```diff
@@ -0,0 +1,68 @@
...
+
+- **Testing limitations:** We know we get the best bang for our buck from [high-level integration tests](https://github.com/DoSomething/rfcs/blob/master/004-testing.md), but separating into services forces us to use mocking extensively because the majority of our application (e.g. users in Rogue, or users & posts in Phoenix) is always an external dependency in any given test! This places a large burden on our intra-service contracts.
+
+- **Knowledge silos:** Finally, this separation reinforces silos (e.g. "you have to talk to Dave about that app" or "that's X application's responsibility, not ours"). We have a smaller team today than when we first started splitting out into a service oriented architecture, and it's important that we can all understand & comfortably work on the whole system.
+
+
+
+## Proposal
+
+We should consolidate our backend services (Northstar, Rogue, and Chompy) into **a single monolithic Laravel application.**
```
aaronschachter: How would this work for the Rogue SQL and Northstar MongoDB databases?  Are you thinking they'd be separate, or migrate one into the other?

DFurnes: For now, I think we'd keep both database connections and specify that the user's model is located on the MongoDB database via [Eloquent's `$connection` property](https://laravel.com/docs/8.x/eloquent#database-connection). Longer-term, I do think it's worth re-evaluating whether MongoDB is worth the overhead it incurs (in terms of technical onboarding & operational complexity) but that's probably another RFC!

aaronschachter: I think that makes sense. Perhaps we should spell it out, or would that make more sense to hold off for a formal technical spec? Chompy has a SQL database used primarily for the RTV import, that I'd imagine would be cleanest to migrate over to the Rogue DB.

014-single-codebase.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Drawbacks/Alternatives
+
+#### Okay, but on the flip side – why _did_ we move from a monolith to (micro)services?
+
+Our Drupal application had a lot of intertwined modules that could be hard to follow, due to Drupal 6's "hooks" based architecture. We wanted to use service boundaries to force a "separation of concerns" that we thought would keep our codebase tidier.
+
+At the time, microservices were the architecture du jour & we followed the success stories of teams many times our size, from Spotify to Google, that promised that smaller services would increase velocity and allow teams to ship faster. At first, we had roughly one team of developers per application ("Rocket" for Phoenix & Northstar, "Bleed" for Rogue, and "O'Doyle" for Gambit & Blink) and it really did increase velocity! (That velocity slowed as a we began to build more cross-cutting features later on.)
+
+We also found that splitting our monolithic application into services allowed us to choose "the best tool for the job", such as using Node.js for our chatbot. (We love PHP at Do Something, but we know it's not appropriate for every task).
```
aaronschachter, line 47: I can't help myself from commenting here that this line makes me daydream of an all JS stack (using Node.js instead of Laravel).. but definitely doesn't make sense as a first pass for Monolith v1

aaronschachter (approved):
> What situations would have been made more difficult if we were all working in a single repository/application?

Having spent a good chunk of time on building complexity into the Rock The Vote (RTV) import this year makes me think about potential added stress if other developers could potentially be blocking deploys per bugs getting merged to master in order to test on QA. These days, we tend to not add too many new backend features, so it seems this would be a rarity.

> Are there other situations where you think that combining Northstar, Rogue, and Chompy would have made work easier?

It'd be great to simplify the RTV import by making Model queries for both users and activity. I'd imagine the Clubs work would have been a little less scattered.

mendelB (approved):
👍   Your reasoning is cogent, I'm on board, and this is very exciting! (I appreciated the illuminating case-study resources!)

--- 

> What situations would have been made more difficult if we were all working in a single repository/application?

Per Aaron's point with Chompy, I suppose we'll have to be more intentional about deploys and blocking features on QA. I feel like we've gotten a lot better at this with feature flagging, and more consistent deploys, but it'll probably take some adjusting to these apps being bunched together I figure?

> Are there other situations where you think that combining Northstar, Rogue, and Chompy would have made work easier?

Your Clubs example resonates viscerally lol. One thing that comes to mind - because I've interacted with our test mock around it recently - is our [Phoenix 'gateway' API to Rogue Campaign & Post endpoints](https://github.com/DoSomething/phoenix-next/blob/aea6446f0c0225364bbfcbd174487b6d8044fa12/routes/api.php#L26-L40), which don't need to exist in such a world.

Another general pain point that comes to mind related to your security & framework upgrades point, is just generally enforcing consistency and convention (code or PR related or otherwise) across the suite, documentation, etc.

014-single-codebase.md
DFurnes: Yeah, being able to use the same language & code across the whole stack is just so incredibly compelling. I'm not sure anything compares with Laravel for "out of the box" productivity in the JavaScript ecosystem, though. (Although I am very eagerly following frameworks like [RedwoodJS](https://redwoodjs.com) that seem to be filling that niche!!)

014-single-codebase.md
DFurnes: I'm starting to sort out a spec for how we'd tackle this work in practice, and will definitely include that there!

014-single-codebase.md
```diff
@@ -0,0 +1,68 @@
...
+
+- It will also **simplify adding behavior that straddles contexts**, like clubs, by replacing network requests with function calls. We'll also be able to ensure better consistency by using more transactional validation.
+
+- Keeping more of our business entities in the same domain allows **easier integration testing**, since we can seed (and otherwise directly reference) data that would currently live across API boundaries, like user IDs & clubs.
+
+- We have less applications to keep up-to-date with security & framework upgrades!
+
+For services that may not make sense to fold into the monolith, like [GraphQL](https://github.com/DoSomething/graphql/) or [Gambit](https://github.com/DoSomething/gambit/), we should still collaborate on them [**as part of a single monorepo**](https://docs.google.com/document/d/1a3e0Wpv9qeRscspdbN4wVR4Chz5oJFb9HH69x2pngSg/edit), side-by-side with our consolidated backend application. This allows us to retain the technical benefits we've gotten from separating out these services, while collaborating in one place.
```
lindsayrmaher, line 35: Can you give me more context on what this means? I read through the monorepo doc but I'm not 100% sure I understand how they would continue to be separate services within the same repo.

lindsayrmaher (approved):

014-single-codebase.md
katiecrane: This includes Aurora too yeah?

014-single-codebase.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+
+## Proposal
+
+We should consolidate our backend services (Northstar, Rogue, and Chompy) into **a single monolithic Laravel application.**
+
+- This will **simplify architectural decisions** & enable us to iterate faster. Instead of sorting new functionality into strict bounded contexts up front, we'll be able to simply place new entities into a single backend context. Architectural decisions will be less binding, as it's easier to refactor within a single codebase than across network boundaries.
+
+- It will also **simplify adding behavior that straddles contexts**, like clubs, by replacing network requests with function calls. We'll also be able to ensure better consistency by using more transactional validation.
```
katiecrane, line 29: This bit definitely is something where we've had long discussions about trade-offs before and I like the benefit we get here!

014-single-codebase.md
katiecrane: +1 I have the same question! I've never really felt that I understood the benefit of the monorepo idea and I'm curious if this would be part of this unified backend work or separate?

DFurnes:
> Having spent a good chunk of time on building complexity into the Rock The Vote (RTV) import this year makes me think about potential added stress if other developers could potentially be blocking deploys per bugs getting merged to master in order to test on QA. These days, we tend to not add too many new backend features, so it seems this would be a rarity.

> Per Aaron's point with Chompy, I suppose we'll have to be more intentional about deploys and blocking features on QA. I feel like we've gotten a lot better at this with feature flagging, and more consistent deploys, but it'll probably take some adjusting to these apps being bunched together I figure?

Yeah, that's definitely a trade-off! I agree that feature flagging & consistent deploys help mitigate this issue, although they're obviously not perfect. One other benefit of consolidating all of these services is that review apps should become a lot more useful (since we can actually spin up a "real" replica of the whole backend, rather than just a little piece of it).

Another trade-off that makes me think of is that test suites for the consolidated backend will take longer to run (since there's just more tests that are living in one place)! I think that's a worthwhile trade-off to make since our "consolidated" integration tests will also be a lot more valuable, but it may be something we need to address down the line (since one of the common complaints from teams with large monoliths is slow test suites).

014-single-codebase.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+
+## Additional Resources
+
+A lot of inspiration for this proposal comes from DHH's [The Majestic Monolith](https://m.signalvnoise.com/the-majestic-monolith/) (2015) and [The Majestic Monolith Can Become The Citadel](https://m.signalvnoise.com/the-majestic-monolith-can-become-the-citadel/) (2020), which detail Basecamp's "majestic" monolithic architecture. They have a small team of 12 developers, and most of their functionality lives in a single codebase (with a few extracted services where necessary).
+
+Another great case study is Segment's [Goodbye Microservices](https://segment.com/blog/goodbye-microservices/) (2018), which explains why they transitioned their famous microservice architecture back to a monolith. They cite improved productivity, testability, and reduced operational overhead.
+
+Finally, in [MonolithFirst](https://martinfowler.com/bliki/MonolithFirst.html) (2015), Martin Fowler argues that projects should start as a monolith and only be broken up once the system's bounded contexts are solidly established. At that point, microservices can be peeled off around the edges.
```
katiecrane, line 68: Sounds like this is the path we did take originally!

014-single-codebase.md
DFurnes: Yes! Aurora is really a front-end though, so maybe I should say "consolidate our Laravel services"...

014-single-codebase.md
```diff
@@ -0,0 +1,68 @@
...
+
+First of all, it takes time! We'd be betting that the time we'd take to rearchitect these applications now would be outweighed by the increase in velocity we'd see as a result. Luckily, we're in a bit of a "lull" at the end of the year & this feels like an opportune time to "reset" and set ourselves up for success in the new year.
+
+We might also find that if we were to grow the team aggressively in the future, we'd once again start tripping over each other in a single codebase. It's hard to predict the future, but it doesn't seem likely at the moment that will be the case.
+
+
+
+## Additional Resources
+
+A lot of inspiration for this proposal comes from DHH's [The Majestic Monolith](https://m.signalvnoise.com/the-majestic-monolith/) (2015) and [The Majestic Monolith Can Become The Citadel](https://m.signalvnoise.com/the-majestic-monolith-can-become-the-citadel/) (2020), which detail Basecamp's "majestic" monolithic architecture. They have a small team of 12 developers, and most of their functionality lives in a single codebase (with a few extracted services where necessary).
```
katiecrane, line 64: >As more and more people come to realize that the chase for microservices ended in a blind alley, the pendulum is going to swing back. The Majestic Monolith is here waiting for microservice refugees. And The Citadel is there to give peace of mind that the pattern will stretch, if they ever do hit that jackpot of becoming a mega-scale app.

Sounds like they'd agree with this plan!

014-single-codebase.md
katiecrane: Yeah I think that makes sense!

014-single-codebase.md
DFurnes: From my perspective, the main benefits we've gotten from splitting out services are allowing them to use a different stack (e.g. using Node.js for Gambit & GraphQL) or scale independently (e.g. how we scale up Northstar and Rogue during SMS broadcasts, but we don't scale up Phoenix).

I _don't_ feel like the "separation of concerns" we get from putting them in separate GitHub repositories is nearly as useful though (and is perhaps even detrimental, since it means you have to make related changes in two or more separate pull requests).

So in cases where we'd want to use a different language or scaling strategy for some part of our stack, I'd still recommend keeping those apps in the same codebase (alongside things that may deploy elsewhere), even if we continue to deploy them separately. That way, for example, a change to GraphQL (a Node.js Lambda app) can happen right alongside the relevant changes to our Laravel backend APIs!

DFurnes: So, a potential monorepo future might look something like this (where each top-level folder is deployed to a different place & perhaps scaled independently, even though they all live in the same GitHub repository):

```
dosomething
⌊backend (backend Laravel codebase)
  ⌊app
  ⌊config
  ⌊database
  ⌊…
⌊graphql (graphql lambda)
  ⌊config
  ⌊docs
  ⌊src
⌊frontend
  ⌊components
  ⌊constants
  ⌊fonts
  ⌊…
```

This also allows services to share common code without a complicated build process (like how we shared Forge as an npm module, or how we pull the GraphQL schema [into Phoenix](https://github.com/DoSomething/phoenix-next/blob/master/schema.json) with a build step).

rapala61 (approved):
Exciting. In my view, this follows a healthy life cycle of our platform. We are coming back to a Monolithic architecture, but with substantially more knowledge to help us evade common pitfalls we found in the previous iteration.

014-single-codebase.md
weerd: I'm sure you're already considering this @DFurnes and I'm playing catch up after being gone for ~4 months... but thoughts on PostgreSQL for the monolith DB? It would support a monolith with services that used a SQL database and might be easier, down the road, to transition MongoDB data into it thanks to it's support for JSON-like data?

014-single-codebase.md
weerd: Alternatively, we could go in the direction of a full Laravel stack, utilizing [Blade components](https://laravel.com/docs/8.x/blade#components) for all our component-ized code needs (which support stuff like props & default props in React but in PHP Blade templates), [Laravel Livewire](https://laravel-livewire.com/) for no-page-refresh updates (get that SPA-feel) and some sprinkling of [AlpineJS](https://github.com/alpinejs/alpine) where needed. And we get server-side rendering (pages and components) cause they're Blade views that get cached along with the great "out of the box" productivity, debugging and security that comes with the Laravel ecosystem. 😈  

I've used a bit of this on my own sites and it's pretty rad; down for whatever the team prefers though! I just wanna make sure we're keeping options open 😚

weerd:
>Another trade-off that makes me think of is that test suites for the consolidated backend will take longer to run (since there's just more tests that are living in one place)! I think that's a worthwhile trade-off to make since our "consolidated" integration tests will also be a lot more valuable, but it may be something we need to address down the line (since one of the common complaints from teams with large monoliths is slow test suites).

Definitely a very good point to consider. I wonder if having the knowledge that test suite runs could take a while, we plan to have the test suite built up in a way that allows easily running specific tests or groups of tests without having to run the entire suite each time during development. Only during deploys or merges to main branch would the entire suite need to be successful. This may be a given, but could be something to just point out so we're all thinking about it going in.

weerd (approved):

DFurnes:
> I wonder if having the knowledge that test suite runs could take a while, we plan to have the test suite built up in a way that allows easily running specific tests or groups of tests without having to run the entire suite each time during development.

Yeah, I think that's definitely important! One of my favorite things about Jest tests is the "watch" mode, by default, only runs tests where files have changed since the last commit. I don't think it's quite as easy to do something like that with higher-level tests, like we write in PHPUnit or Cypress, but we can definitely get in the habit of running smaller segments/groups of tests!

And, for PHPUnit in particular, I think we also have a lot of performance wins we could probably find – our whole Rogue test suite finishes in a speedy 15 seconds on my machine, whereas Northstar's suite takes almost 2 minutes (largely, I think, because it can't use database transactions to easily roll things back between test cases).

014-single-codebase.md
DFurnes: I think that's a great idea! PostgreSQL seems like a really good fit with its JSON column support, and we could consider using Aurora Serverless to cut down costs a bit more along the way. Plus, we already have a lot of in-house expertise using PostgreSQL on the data side of the house! 🤓

014-single-codebase.md
DFurnes: It does look incredibly slick, and makes a ton of sense for a smaller team like ours.

From a practical perspective, I'm wary of anything that'd involve too much rewriting (either from Laravel to Node.js, or from React to Livewire/Alpine), since that leaves us potentially "spinning our wheels" for a long time. If we were to consider something like that, I think we'd need a good plan on how we'd manage that transition effectively.

weerd: Yep totally agree @DFurnes. I mostly just brought it up so as to keep the TALL (Tailwind, AlpineJS, Livewire & Laravel) Stack in consideration and if it could serve our purposes for some apps (or depending on how it comes together as a single repo), especially if some of our stack is already heavy in some parts with Laravel.

I think for Phoenix and the UI for the main website it makes sense to continue down the NodeJS path and more specifically ReactJS & NextJS, but I have thought in passing that the TALL stack could work great for our more backend heavy functionality like user "Activity" (what Rogue handles). I think the Rogue Admin UI in particular, would be a solid candidate to switch it to componentize it with Blade components and use AlpineJS, Livewire and Tailwind for styling.

But ➕ to needing to have a good plan and discuss how to manage it and even _if_ it would makes sense to go that route! Just sharing thoughts for now 👯‍♂️

PR #15, opened 2021-03-30
Refactor campaign transactional emails

aaronschachter:
This pull request introduces the "Customer.io Transactionals" RFC:

> We should refactor sending campaign activity transactional emails to use Customer.io's App API instead of the Track API to avoid maintaining a Customer.io profile for a member to send transactional emails solely.

## Questions

With limited resources, I'm seconding guessing whether doing this work for the sake of sending transactionals to unsubscribed members is worth the effort. 

We don't want to maintain a customer.io profile for members who unsubscribed to send the transactional, and it for sure made sense to refactor Northstar to use the App API to send transactional password reset/updated confirmation emails. 

Maybe a minimal footprint approach could be adding a new email subscription topic specifically for members who want to receive campaign transactionals (opted in by default) -- so campaigns/marketing can still effectively manage all transactionals without any code/engineering support.


015-customer-io-transactionals.md
```diff
@@ -0,0 +1,44 @@
+In 2021, we began to [only maintain Customer.io profiles for active members who have opted in to receive email and/or SMS promotions](https://www.pivotaltracker.com/epic/show/4721712). The [workflows we use in Customer.io](https://customer.io/visual-workflow-builder/) are convenient because they provide a no-code way to send transactional email and/or SMS messages -- but it does require that a member has a Customer.io profile in order to receive the transactional message. It's also a bit brittle (e.g. -- an admin could accidentally delete the workflow that sends our SMS compliance message if someone opts-in to receive texts via web)
```
DFurnes, line 1: The ability to accidentally disable our SMS compliance message definitely seems like a big risk factor & a good argument for moving critical messaging flows like this into code. 👍

DFurnes:
> With limited resources, I'm seconding guessing whether doing this work for the sake of sending transactionals to unsubscribed members is worth the effort.

Yeah, I can see good arguments either way! The transactional API seems like a really good approach for "critical" notifications, like password resets or SMS compliance messaging. For folks who are unsubscribed, though, I wonder if other transactional messages are something they'd still expect to get...

> Maybe a minimal footprint approach could be adding a new email subscription topic specifically for members who want to receive campaign transactionals (opted in by default) -- so campaigns/marketing can still effectively manage all transactionals without any code/engineering support.

That makes a lot of sense to me, and seems to mirror email/notification settings for a lot of other services. I'm into it!

PR #16, opened 2021-04-29
Add RFC for Legacy Code Workflow.

weerd:
This pull request introduces the "Legacy Code Workflow" RFC:

> We should establish a transition workflow for how we refactor and swap out legacy code components or encompassed-logic.

### DIFF
--- a/16-legacy-code-workflow.md
+++ b/16-legacy-code-workflow.md
@@ -1,0 +1,49 @@
+# Legacy Code Workflow
+
+```
+Howdy team! I was chatting with Dave about how to better handle how we refactor and sometimes replace old components/encompassed-logic in our repositories. A good example is what I’m currently working on, with some major changes to the EmailSubscriptions component that we see in the user profile. There’s lots of new functionality and visual changes so it made sense to me to create a new component, pull what I needed from old and add new code. This also allowed me to work on it and now break the existing component that is deployed. However it could be confusing if not everyone is aware it’s being changed out and when to start using the new one.
+
+A suggestion I had was to first rename the existing component and prefix it with Legacy.... Not only would this highlight to other devs that it’s a component on the way out and consider using the new one if it’s available, but it would also help the dev working on the changes to scour the codebase and find all iterations of the old component to then rename. It would provide a better overall view of everything that’s going to be effected, etc.
+3:32
+I initially just named the component something different NewsletterSubscriptions so it would not conflict, but I feel the above approach would be better and also immediately highlight the transition happening in the codebase to the rest of the team. So if you see LegacyEmailSubscriptions, that means it’s something on the verge of deprecation (or should be deprecated) and instead look for EmailSubscriptions which would be the new component!
+```
+
+We should establish a transition workflow for how we refactor and swap out legacy code components or encompassed-logic.
+
+**Discussion:** <https://github.com/DoSomething/rfcs/pull/16>
+
+## Problem
+
+When we are replacing an established component with a new component or replacing encompassed code logic in a function or class, the process may take multiple pull requests or steps to complete to avoid issues with the legacy code currently in use.
+
+As a developer proceeds through the process, they need a better way to transition replacing the existing code without causing downtime or confusing other developers as to which code approach should be used and which one is being deprecated.
+
+## Proposal
+
+To avoid downtime and remove confusion to developers the proposal is to begin the process by renaming the existing component/class/method/function and prefixing it with `Legacy*`. This would be the first pull request when tackling the task.
+
+For example, a component called `EmailSubscriptions` would be renamed to `LegacyEmailSubscriptions`, and all instances where it is used would be renamed as well in the first pull request for review.
+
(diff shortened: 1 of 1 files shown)


DFurnes (approved):
Makes a ton of sense to me!

16-legacy-code-workflow.md
```diff
@@ -0,0 +1,41 @@
...
+
+To avoid downtime and remove confusion to developers the proposal is to begin the process by renaming the existing component/class/method/function and prefixing it with `Legacy*`. This would be the first pull request when tackling the task.
+
+For example, a component called `EmailSubscriptions` would be renamed to `LegacyEmailSubscriptions`, and all instances where it is used would be renamed as well in the first pull request for review.
+
+This would immediately highlight to other developers that the encompassed code is being deprecated and should potentially utilize the new version (ideally named the same, but without the `Legacy*` prefix).
+
+If the new encompassed code is renamed to something different, the legacy code with the `Legacy*` prefix could have a code comment indicating the new name of the code that is replacing it.
+
+_Note:_ When available, it would also be highly recommended to include a `@deprecated` in the code block for the legacy code.
```
DFurnes, line 23: I love how VS Code (and probably other editors, too) will then give you a heads up when you try to use that deprecated method or component by crossing out the name! Such a handy visual aid. 〰️

16-legacy-code-workflow.md
```diff
@@ -0,0 +1,41 @@
...
+
+However, as mentioned, this initial work would help familiarize the developer with the changes and the extent of what needs to be replaced/edited.
+
+## Alternatives
+
+Within a team discssion on Slack, the wise puppet-master himself and DoSomething alumnus (RIP), Aaron, suggested maybe placing the renamed legacy code within a `legacy` directory that could afterwards be deleted when the code is finally removed.
+
+The benefit of this would be that it would be one ideal spot to contain all legacy code.
+
+This could be a useful idea, but not all code would be easily moved from one directory to another directory. It would be easier for component code, but for classes, methods or functions it would be more ideal and possibly required for them to stay in their original locations until the new code is available to replace them. It would also add an additional step of needing to move legacy code all to one spot along with renaming them.
```
DFurnes, line 41: Yeah, my first instinct was to want to make a legacy directory for these too (just like we do [in backend code](https://github.com/DoSomething/northstar/tree/main/app/Http/Controllers/Legacy)!), but I agree that it becomes a bit more of an ordeal with how relative paths work in our JavaScript codebases.

16-legacy-code-workflow.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Proposal
+
+To avoid downtime and remove confusion to developers the proposal is to begin the process by renaming the existing component/class/method/function and prefixing it with `Legacy*`. This would be the first pull request when tackling the task.
+
+For example, a component called `EmailSubscriptions` would be renamed to `LegacyEmailSubscriptions`, and all instances where it is used would be renamed as well in the first pull request for review.
+
+This would immediately highlight to other developers that the encompassed code is being deprecated and should potentially utilize the new version (ideally named the same, but without the `Legacy*` prefix).
+
+If the new encompassed code is renamed to something different, the legacy code with the `Legacy*` prefix could have a code comment indicating the new name of the code that is replacing it.
```
mendelB, line 21: Another use case that springs to mind (and is tangentially related to points below) is if we move the new component to a more appropriate location (from root to /utilities). This could be an opportunity to add another comment, or better yet, first, move the legacy component to its new location in the first place 🤔

16-legacy-code-workflow.md
mendelB: Could we consider a `Deprecated` prefix to components still in use that we intend to deprecate, but do _not_ intend to rebuild a new version of? (I doubt this is a common scenario... but just a spare thought.)

mendelB (approved):
Makes great sense! 👍

16-legacy-code-workflow.md
weerd: oooo, I dig this idea. I think the code block `@deprecated` would essentially do that, but I could maybe see scenarios where we don't have code blocks (probs should) and deprecating something.... thoughts?

16-legacy-code-workflow.md
weerd: Ya, great point!

16-legacy-code-workflow.md
```diff
@@ -0,0 +1,41 @@
...
+
+_Note:_ When available, it would also be highly recommended to include a `@deprecated` in the code block for the legacy code.
+
+An added benefit of renaming the legacy code with the `Legacy*` prefix is that it would encourage the developer to also scour the codebase to find all iterations of the old, legacy code, rename it and thus highlight all instances where it is used so they familiarize themselves with where all the changes will be taking place. This would provide a better overall view of everything that will be affected by the changes.
```
katiecrane, line 25: I love this benefit!

katiecrane (approved):
I have definitely been confused about which is the new vs. old version of something and really looking forward to the clarity this will bring!

PR #18, opened 2021-05-04
Add RFC for InertiaJS.

weerd:
This pull request introduces the "InertiaJS" RFC:

> We should implement the use of [Inertia JS](https://inertiajs.com/) to simplify the development of our frontend user interfaces for our administrative backend.

### DIFF
--- a/018-inertiajs.md
+++ b/018-inertiajs.md
@@ -86,6 +86,4 @@ While this sounds useful, the React Admin framework, as mentioned, relies heavil
 
 That being said, there could be other React admin libraries that could supply more basic components and alleviate the need to build our own.
 
-I think it would benefit the team to potentially build custom components that could be shared with both the Inertia adapted Laravel backend and the member facing frontend UI, but all the endpoints is something the team could discuss further and not dependent on the core of this RFC t
-
-o decide on implementing the use of Inertia.
+I think it would benefit the team to potentially build custom components that could be shared with both the Inertia adapted Laravel backend and the member facing frontend UI, but all the endpoints is something the team could discuss further and not dependent on the core of this RFC to decide on implementing the use of Inertia.
--- a/018-inertiajs.md
+++ b/018-inertiajs.md
@@ -1,0 +1,56 @@
+# Inertia JS
+
+We should implement the use of [Inertia JS](https://inertiajs.com/) to simplify the development of our frontend user interfaces for our administrative backend.
+
+## Problem
+
+We are in the process of consolidating and merging multiple backend services into a single codebase. Each of these services had their own frontend administrative interfaces, and we will need to unify them as well.
+
+As a team, we have found that moving to a component-based architecture for the frontend using [React](https://reactjs.org/docs/getting-started.html) and [Tailwind](https://tailwindcss.com/docs) has helped us re-use code more efficiently and made it easier to maintain existing features and develop new ones.
+
+However, to implement a component-based architecture for the administrative frontend of our consolidated backend would require a lot of work to expose numerous new API endpoints and connections that React could hook into, just to enable the administrative interface.
+
+It would be ideal if we could maintain the backend code and have it communicate easily with the frontend without having to add a lot of confusing duplicate code or connectors and endpoints that would then need to be maintained.
+
+This is where Inertia JS would excel at providing the ideal solution; it would allow us to enable a fully client-side rendered frontend and provides the glue to connect it to our existing server-side Laravel PHP code.
+
+The updates to connect the pieces are minimal and gets the team up and running to develop our administrative frontend utilizing React components and Tailwind.
+
+## Proposal
+
+Implement the use of [Inertia JS](https://inertiajs.com/) in Northstar for our administrative interface.
+
+The overall aim is to reduce maintenance for our team, particularly when adding or updating new features to our administrative interfaces.
+
+Inertia will allow us to use the best of both worlds; a client-side rendered interface with reusable React components and server-side business logic utilizing the robust and familiar Laravel framework without worrying about additional code to pass data to the React components.
+
+There will be an initial ramp up to build out the core components for our administrative interfaces. However, we can build on the experience we have obtained from component-izing our user facing frontend. We also already have some components from our activity service that have been built using React, and we could easily update those to receive data props from Inertia. Once we have an established baseline of admin components most of the admin pages will be easier to build out and/or update.
+
+The following is a proposed approach for implementing Inertia JS in Northstar:
+
+- Add Inertia JS to Northstar and set it up so it works with our existing code.
+- Implement a new main layout for the admin interface.
+- Consolidate all routes into Inertia routes that will pass data as props.
+- Build out the interface using React components that utilize the data passed via the Inertia routes.
+
+## Drawbacks
+
+For the most part there do not appear to be any significant drawbacks with implementing Inertia JS. It seems like the right "happy middleground" that would let us maintain most of our existing Laravel backend work, yet provide the flexibility to use component-based interface elements built in React.
+
+One potential drawback is that the team would be required to update and create admin interface elements; especially given how specialized some of our needs can be at times. However, I imagine that after building a number of initial admin interface elements in React as components, the level of effort would drop as components can be re-used.
+
+Another potential drawback is that Inertia just has not been around for that long. The [Inertia adaptor for Laravel](https://github.com/inertiajs/inertia-laravel) is currently tagged at version `v0.4.1` and the [Inertia adaptor for React](https://github.com/inertiajs/inertia/blob/master/packages/inertia-react/package.json) is currently at version `v0.5.12`.
+
+That being said, it has had a lot of support within the Laravel community and a lot of recent commit activity. The documentation is very thorough and it even has an entire [live example demo site](https://inertiajs.com/demo-application) built using Laravel, InertiaJS and VueJS with the code available as an [open source repository](https://github.com/inertiajs/pingcrm). The community has even contributed versions of the demo site using other server-side and client-side languages/frameworks, including one using [Laravel, InertiaJS and ReactJS](https://github.com/Landish/pingcrm-react).
+
+## Alternatives
+
+There are a couple of alternatives we also considered, but with either one it would potentially mean either less flexibility, or lots of overhead and more maintenance to enable our existing backend code to communicate with the frontend.
+
+### Laravel Nova
+
+// @todo
+
+### React Admin
+
+// @todo
--- a/018-inertiajs.md
+++ b/018-inertiajs.md
@@ -61,4 +61,17 @@ Nova does offer CLI generators to create custom tools within the framework that 
 
 ### React Admin
 
-// @todo
+[React Admin](https://marmelab.com/react-admin) is an open-source framework for building admin applications in React with data provided by REST/GraphQL APIs.
+
+The framework looks like is has a number of useful components to help build out an administrative interface. The main benefit React Admin provides is that the team would not have to do much work building out components for an administrative interface.
+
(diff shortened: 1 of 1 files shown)


018-inertiajs.md
```diff
@@ -0,0 +1,81 @@
...
+
+That being said, it has had a lot of support within the Laravel community and a lot of recent commit activity. The documentation is very thorough and it even has an entire [live example demo site](https://inertiajs.com/demo-application) built using Laravel, InertiaJS and VueJS with the code available as an [open source repository](https://github.com/inertiajs/pingcrm). The community has even contributed versions of the demo site using other server-side and client-side languages/frameworks, including one using [Laravel, InertiaJS and ReactJS](https://github.com/Landish/pingcrm-react).
+
+## Alternatives
+
+There are a couple of alternatives we also considered, but with either one it would potentially mean either less flexibility, or lots of overhead and more maintenance to enable our existing backend code to communicate with the frontend.
```
katiecrane, line 56: I'd love to hear more about what the alternatives were!

katiecrane: OH I see this continues below don't mind me

018-inertiajs.md
```diff
@@ -0,0 +1,81 @@
...
+
+Nova does offer CLI generators to create custom tools within the framework that could support our business needs. The downside is that these tools are generated as Vue components, which would require our developers to learn [Vue JS](https://vuejs.org/) and context switch between Vue JS for administrative tools and React JS for member facing tools. While the need for custom tools may be low, it would imply additional maintenance overhead in a client-side framework the team is not familiar with.
+
+### React Admin
+
+[React Admin](https://marmelab.com/react-admin) is an open-source framework for building admin applications in React with data provided by REST/GraphQL APIs.
+
+The framework looks like is has a number of useful components to help build out an administrative interface. The main benefit React Admin provides is that the team would not have to do much work building out components for an administrative interface.
+
+However, the framework is dependent on having established API endpoints for all the resources it will render. This would require the team to invest time and effort into building out endpoints for all backend resources and maintaining them over time.
```
katiecrane: Don't we already do this? I don't think there's really anything from the former Rogue at least that isn't also surfaced in the API but maybe I'm not thinking of the right example here!

018-inertiajs.md
weerd: It's the alternative below Laravel Nova or the React Admin framework :)

018-inertiajs.md
weerd: I'm not 100% sure everything from Rogue was surfaced for the Rogue admin. Not all interfaces in Rogue admin were React. But now that we're merging everything into Northstar, it's more than just the routes in Rogue... Aurora and Chompy too, so we'd have to expose API endpoints for all of them as well and maintain them if not already available and transitioned over.

It would significantly reduce our efforts in maintaining code if all API endpoints (from all services combined) in Northstar are only endpoints needed by Phoenix or other dependent external services, instead of maintaining those AND endpoints just for an administrative frontend. It would also just make the code more readable to know that normal web routes (thanks to Inertia) supply the data to the React frontend, and API endpoints are solely for external apps!

018-inertiajs.md
weerd: GitHub didn't show me you already commented on seeing the part that continues below 😅

018-inertiajs.md
katiecrane: I'm into the idea of Inertia but React Admin seems cool too so I want to make sure I understand why it was ruled out, so I have a few more curiosities! 

Agreed that not all interfaces in Rogue were React and that maybe not everything was surfaced for Admins, but I think everything was surfaced via the API, whether we actually used it for the front-end or not. Chompy I am less sure about and could see being a pain point for sure. But we know everything in Aurora was available via the API since it was a separate app from Northstar. 

I see how this approach could cause extra API maintenance like if there are things we need to surface to admins that Phoenix never needs to know about - is that the sort of thing you're thinking about? Or is there a historical example where maybe using the Northstar API for both Phoenix and Aurora caused issues?

weerd: > I see how this approach could cause extra API maintenance like if there are things we need to surface to admins that Phoenix never needs to know about - is that the sort of thing you're thinking about?

Ah yes, moreso this! Sorry if I wasn't super clear and if desired, I can amend the text to make this point clearer. 😅 

If everything is an API endpoint for both the admin frontend and the external services, there will be a number of endpoints the external services do not care about (only used for the admin frontend) and plenty where there will be overlap which both would use. This could cause problems in itself and needing to constantly feature-flag or version control endpoints if we want to change something up for one but not the other. Having an external dependency that could break when changes are made to an endpoint adds extra overhead and complexity.

Inertia essentially lets us keep all our `routes/web.php` for our React-built admin app and `routes/api.php` for any external services, which feels like a nice separation knowing exactly where to go for what you need to do.

But, as I mention briefly in the "Combo" section, just because we use Inertia does _not_ mean we cannot use a React UI library similar to React Admin (but just UI components, without it doing its own requests for data) instead of rolling our own. I would like to explore rolling our own components initially to see how feasible it is and if we could have a collection of basic components that both Northstar Admin and Phoenix could utilize (new Forge? 🤔 ).

katiecrane: Thanks Diego - I totally see where you're coming from more clearly now! Particularly liked the example where maybe we want to change something for one but not the other - having them separated definitely avoids that! I think it couldn't hurt to pop some of this reasoning into the main RFC itself but up to you!

katiecrane (approved):
No further questions! I like this approach and that we still have the "combo" option if there are other useful libraries out there.

018-inertiajs.md
weerd: Sounds good! I'll amend the RFC and add this into it. Appreciate you questioning it, since I hadn't realized it wasn't super clear on that point. 🙌

katiecrane: Yeah yeah! I think I was coming at it from a different angle - thanks for taking the time to really explain it to me, much appreciated!

018-inertiajs.md
```diff
@@ -0,0 +1,81 @@
...
+
+However, I am cautious that how our Eloquent models have evolved and the different relationships we have established between resources, combined with the merging of services may not completely follow the Laravel approach and we could run into numerous headaches.
+
+I also imagine we would have to wait until we complete the unified database and switch off MongoDB completely before we could move to use Nova, to avoid issues since the MongoDB Eloquent driver is a community contribution and not officially supported by Laravel.
+
+Nova does offer CLI generators to create custom tools within the framework that could support our business needs. The downside is that these tools are generated as Vue components, which would require our developers to learn [Vue JS](https://vuejs.org/) and context switch between Vue JS for administrative tools and React JS for member facing tools. While the need for custom tools may be low, it would imply additional maintenance overhead in a client-side framework the team is not familiar with.
```
DFurnes, line 68: I didn't realize Nova was so tightly coupled to Vue! That's definitely a big downside for our needs.

weerd: Ya 😞  I wonder if they'll ever expand to other reactive frameworks but unless there's a high demand I doubt it. I do like VueJS but agreed that it's not in-line with DoSomething's needs.

018-inertiajs.md
DFurnes: This is a great discussion, and definitely worth writing up a bit more on in the RFC! 🙌 

I think we generally created non-API admin panels (like the "create campaign" and "create action" forms) when it seemed like it'd be too much effort to wire all this up to an API endpoint & GraphQL mutation. Something I'm excited to experiment with is Inertia's support for form submission & [validation](https://inertiajs.com/validation), which would probably have made these two forms much less daunting to build in React.

018-inertiajs.md
```diff
@@ -0,0 +1,81 @@
...
+- Build out the interface using React components that utilize the data passed via the Inertia routes.
+
+## Drawbacks
+
+For the most part there do not appear to be any significant drawbacks with implementing Inertia JS. It seems like the right "happy middleground" that would let us maintain most of our existing Laravel backend work, yet provide the flexibility to use component-based interface elements built in React.
+
+One potential drawback is that the team would be required to update and create admin interface elements; especially given how specialized some of our needs can be at times. However, I imagine that after building a number of initial admin interface elements in React as components, the level of effort would drop as components can be re-used.
+
+Another potential drawback is that Inertia just has not been around for that long. The [Inertia adaptor for Laravel](https://github.com/inertiajs/inertia-laravel) is currently tagged at version `v0.4.1` and the [Inertia adaptor for React](https://github.com/inertiajs/inertia/blob/master/packages/inertia-react/package.json) is currently at version `v0.5.12`.
```
DFurnes, line 42: Do we know about other companies/organizations that are using Inertia in production? It would make me feel a lot more comfortable picking up a newer tool if, say, some of our non-profit friends were already using it!

018-inertiajs.md
katiecrane: Ah those are great examples, that makes a lot of sense! I thought the form submission & validation stuff that comes with it is super cool too!

018-inertiajs.md
weerd: I added some additional info regarding this!

weerd:
@katiecrane @DFurnes I updated the RFC to include the info we discussed that would be helpful regarding more on caveats of going with solution like React Admin that requires maintaining a comprehensive endpoint API, along with a blurb about other companies using Inertia JS. 😀

018-inertiajs.md
mendelB: Ooh yeah, this is disappointing, it seemed like such an exciting potential avenue for us from that presentation at Laracon!

018-inertiajs.md
```diff
@@ -0,0 +1,97 @@
...
+
+### React Admin
+
+[React Admin](https://marmelab.com/react-admin) is an open-source framework for building admin applications in React with data provided by REST/GraphQL APIs.
+
+The framework looks like it has a number of useful components to help build out an administrative interface. The main benefit React Admin provides is that the team would not have to do much work building out components for an administrative interface.
+
+However, the framework is dependent on having established API endpoints for all the resources it will render. This would require the team to invest time and effort into building out endpoints for all backend resources that would be administered via a frontend UI and maintaining all the endpoints over time.
```
mendelB, line 76: Is the reverse at all a concern -- do we have any API-based resources from Rogue that we'd need to build new route/controllers for to enable Intertia? (And/or perhaps that's an upside if philosophically we'd rather have classic web routes for non-external-based needs?)

mendelB (approved):
Thanks for the write-up! I think you present a compelling case here, and I'm excited to see how Intertia can help us build out a more robust admin interface reliably and accessibly!

018-inertiajs.md
weerd: I didn't do a complete review of the Rogue, now Northstar code regarding this, but given a lot of the resources were available to the front-end of the Rogue Admin I think any API-only resources were very few!

DFurnes: As far as I know, our ~Rogue~ Northstar admin UIs rely on the `v3/` API endpoints (via GraphQL) for most interfaces so we would have to create new Inertia controllers for all of those interfaces.

weerd: Ah right, good call! I was wrong! 😅 

But that's for resources now native to Northstar for the most part so I think it'd be pretty straightforward and eliminate a roundtrip to a different service for a resource that would be used for the UI. I think in the long run it would reduce the overhead of our GraphQL service if GraphQL only cared about and needs to handle shared service resources and not for resources used mostly for Admin purposes (oversimplifying it, but I hope that makes sense!)

And we should still be able to use the React components built for the`v3/` endpoints; just a change in what's feeding them the data 😄

